PR #4, opened 2010-11-17
Add user-Agent to header because some tracker need it.

dufoli:
Bug have been reported by sleepwalker on forum.


usikpavel:
In the latest version (1.0.22) I found that User-Agent header doesn't exist. So, I suggest a fix:
`client.DefaultRequestHeaders.Add ("User-Agent", VersionInfo.ClientVersion);`
This is line **88** in file **MonoTorrent.Client.Tracker/HTTPTracker.cs**

PR #9, opened 2011-08-19
add ConnectToWebSeedsImmediately setting

proff:


brunolemos:
@proff could you share it? thanks :)

I'm looking forward a torrent lib that works on windows phone / universal apps.

alanmcgovern:
If it's annoying, submit a PR with a fix, or at least say what the annoyance is :P

proff:
@diantahoc @brunolemos I'm sorry, but I didn't have time to implement features that were not necessary for my project before the great American liberator, along with his European slave, came to my house and liberated me from the right to a home, the right to land, the right to freedom of choice, the right to freedom of speech and the right to life.

proff:
@alanmcgovern ..He said after 8 years. In my case, I needed a solution capable of seeding tens of thousands or even hundreds of thousands of torrents without performance degradation. At the same time, obviously, there are much fewer active torrents. Because this is server-side seeding for an online distribution system. The architecture as a whole was completely unsuitable, in the end.

Let's try from a clean slate and start from bugs. #723

PR #14, opened 2012-01-12
Changing how the webseed is consumed

victorarias:
I had a problem with the clearing of the seed list because they are never accessible again, and they may be necessary if for any reason the webseed peer disconnect

The src/MonoTorrent/MonoTorrent.Client/Modes/Mode.cs change should fix the problem.


samhocevar:
Hi. This patch contains a lot of unrelated modifications to project files, and also changes the indentation between spaces and tabs such that reading the exact diff is hard, so I cannot apply it as is.

I understand this is very, very old, and you may no longer be interested in getting this applied. If that’s the case, I can clean up the patch myself.

alanmcgovern:

PR #37, opened 2014-05-12
Fixed an issue with the udp tracker spec implementation.

joshmackey:


samhocevar:
This patch looks mostly good, but it also fixes the indentation in `UdpTracker.cs`, which makes the diff unreadable and hard to review. I agree the indentation should also be fixed, but maybe in a separate patch. Any chance you could rewrite the patch? Otherwise I can do it, but in that case you will not appear as the committer.

alanmcgovern:

Issue #33, opened 2013-11-13
TorrentManager should not call UpdateChunks on the engine's rate limiter


hhowe29:
We have had ongoing problems with global rate limits not slowing torrents down. To reproduce

1- start a torrent download with these settings
EngineSettings.MaxDownloadSpeed = 10*1024
TorrentSettings.MaxDownloadSpeed = 0; //unlimited.

Observe the torrent download speed on a high throughput network. We see it skyrocket and stay consistently above 10k

We traced this down to the fact that the TorrentManger calls UpdateChunks on the engine's rate limiter by way of a rate limiter group. However, it passes the torrent managers max rate of 0, and the torrent's observed down rate. It does this in addition to calling UpdateChunks on its own rate limiter.

The call from the torrent manager overwrites the chunk calculation that the engine last did, and sets the unlimited field in the limiter to true. That in turn kills the global limiter's ability to throttle the speed. We decided that the torrrent manager shouldn't be calling the global limiter's UpdateChunks, and fixed it by adding some ugly flags that tell the torrent manager to skip it. It does still call TryProcess on it though.

I will try to get this pushed to our forked monotorrent repo. I would invite you do pursue a cleaner implementation than what we did though.

alanmcgovern:
It seems to work fine in the latest build

PR #39, opened 2014-07-14
Various cleanup plus a fix for bug #33

hhowe29:
Commit 9a6c9eb contains a bug fix to global rate limiting that others might find handy. The other commits are of marginal value to the main repo.

Credit for the bug fix goes to coworker Clayton Stangeland.


alanmcgovern:

PR #41, opened 2014-07-22
Added .gitignore and exposed DiskManager::CloseFileStreams

rastating:
The CloseFileStreams method was marked as internal, but I have found it useful to have access to in the calling assembly however, as when changing state from downloading to seeding it doesn't flush the stream, so I have to do the following to work around it:

```
        private void TorrentStateChanged(object sender, TorrentStateChangedEventArgs e)
        {
            if (e.OldState == TorrentState.Downloading && e.NewState == TorrentState.Seeding)
            {
                e.TorrentManager.Stop();
                e.TorrentManager.Engine.DiskManager.CloseFileStreams(e.TorrentManager).WaitOne();
                e.TorrentManager.Start();
            }
        }
```

In TorrentManager::RaiseTorrentStateChanged there is the following comment:

> Whenever we have a state change, we need to make sure that we flush the buffers.
> For example, Started->Paused, Started->Stopped, Downloading->Seeding etc should all 
> flush to disk.

But to do this, we need the CloseFileStreams method to be exposed. I think it would probably be better to implement the above code snippet into RaiseTorrentStateChanged itself, but not sure if you want that to be implemented outside it or not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+build/
+*obj/
+*.sln
+*.suo
--- a/src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
@@ -146,7 +146,7 @@ namespace MonoTorrent.Client
 
         #region Methods
 
-        internal WaitHandle CloseFileStreams(TorrentManager manager)
+        public WaitHandle CloseFileStreams(TorrentManager manager)
         {
             ManualResetEvent handle = new ManualResetEvent(false);
 


samhocevar:
Is `CloseFileStreams` the best choice her? It seems to offer slightly less control than my own workaround, which is to manually create the `DiskWriter` and keep a handle to it:

        DiskWriter disk_writer = new DiskWriter();
        ClientEngine client_engine = new ClientEngine(settings, disk_writer);

Then the following callback:

    manager.TorrentStateChanged += delegate (object o, TorrentStateChangedEventArgs e)
    {
        if (e.NewState == TorrentState.Seeding)
        {
            foreach (TorrentFile f in e.TorrentManager.Torrent.Files)
                disk_writer.Close(f);
        }
    };

alanmcgovern:
The core should probably flush the streams at the appropriate time. Want to submit a PR that makes MonoTorrent flush?

PR #46, opened 2015-05-13
Fix duplicate peer bug when peers first added after announce

rcawston:
Prevents Available peer count from fluctuating wildly after an announce.
peerId is not set for newly returned peers immediately after Announce() is called.
Previously, this resulted in equal peers being identified as different peers.
With this change, if peerId isn't set for EITHER old or new peer, comparison by host is performed.


samhocevar:
Thanks, this fix makes perfect sense.

PR #47, opened 2015-06-23
Update the SlidingWindow picker to allow you to properly use it in a …

chen78:
…piece picker chain.

svn path=/branches/bitsharp-0.90/; revision=157772


samhocevar:
Apparently this change was already merged into master.

Issue #53, opened 2017-02-15
Peer download rates are incorrect resulting in slow/capped peer download speeds.


TheOriginalWinCat:
There is a bug in the SpeedMonitor Tick() that results in all connected peers have a calculated download speed of 0kb/s.

This is caused by Tick() only updating if over 800ms have elapsed since last called, but the logic that calls Tick() for each peer runs every 500ms meaning the time since lastUpdated is never greater than 800ms.

This is the root cause of the poor download speeds, as the download speed are used to work out how many active requests to have for a given peer, since mono thinks it's downloading from all peers at 0kb/s it only sends one request at a time, resulting in single 16kb requests per a peer.

Thus limiting the speed of each connection to 1.64 Mbit/sec for a RRT of 80ms.
Making it basically impossible to get decent speeds unless seeding from many peers.

After fixing the logic in Tick() mono correctly scales up the active request rates per a peer as intended removing the download speed handicap.

samhocevar:
Fixed in PR #54.

PR #54, opened 2017-02-15
Fix SpeedMonitor Tick() logic. Fixes slow peer download speed #53

TheOriginalWinCat:
Only update the lastUpdated value if we actually perform a update, otherwise if Tick() is called at a fixed rate lower than 800ms the internal average is never calculated. #53


samhocevar:
Good catch! This makes perfect sense, thanks.

PR #80, opened 2018-08-09
Fix DHT

aaronsace:


alanmcgovern:
This doesn't look like it fixes anything. Am i missing something?

aaronsace:
Mainly, it was trying to write to the console and it sometimes sat with no peers at all when using DHT without tracker. 

Also, the client didn't have access to the number of nodes.

PR #84, opened 2018-12-07
fixed: opening source file in non-Western language OS causes source file corruption

imays76:


alanmcgovern:
I changed it to a unicode literal, `\u0192`.

Thanks!

PR #91, opened 2019-05-04
May the fourth be with you!

alanmcgovern:
So I was kinda bored recently and was curious to see how things
would look if MonoTorrent were rewritten to use async/await.

It turns out that not only does it look awesome, but it *is*
awesome! Gone are the hundreds of lines of carefully constructed
BeginX/EndX calls, each with dozens of lines of complicated and
hard to understand error handling.

Gone is the impenetrable flow of callbacks when negotiating an
encrypted connection.

Gone is the magic knowledge you need to correctly allocate/free
buffers. Thanks to 'synchronous' approach offered by async/await,
it's trivial to see where a buffer is allocated and where it needs
to be freed. I even fixed an issue where buffers were GC'ed instead
of being reused because it was now possible to sanely track.

I did write a custom awaiter for the MainLoop because why not. It
does make some code relatively easy to understand now. The threading
concerns are a lot simpler.

The DHT code was also rewritten to use asnyc/await too.

This is all barely tested, but sure it's opensource. If it doesn't
work then go ahead and fork it and fix it. Upstreamed patches would
be nice, but i probably won't be monitoring this.

Enjoy this gif of ubuntu downloading as fast as my connection allows. ~2GB of data, using ~22MB ram, in under 60 seconds makes it at 35MB/sec overall average:

downloading


alanmcgovern:
#yolo

PR #92, opened 2019-05-04
Update LocalPeerManager.cs to fix Local peer discovery with multiple NICs

Kradxn:
Previously local peer discovery was only working on the main network interface, this should fix it for having multiple e.g. WiFi and lan


alanmcgovern:
Great stuff! Thanks for the patch!

PR #98, opened 2019-07-24
Dht updates

alanmcgovern:


src/DotNetCoreSample/Program.cs
```diff
@@ -93,7 +93,7 @@ private static async Task StartEngine()
             DhtEngine dht = new DhtEngine (dhtListner);
             engine.RegisterDht(dht);
             dhtListner.Start();
-            engine.DhtEngine.Start(nodes);
+            engine.DhtEngine.StartAsync(nodes);
```
uygary: This needs await.

uygary: This needs await.

alanmcgovern: I deleted this file completely now :)

PR #99, opened 2019-07-31
Fix some async and cleanup issues

alanmcgovern:
WIP


alanmcgovern:
I think these parts are definitely good to go in!

PR #100, opened 2019-07-31
1、bug fixed--memory leaking caused by SocketAsyncEventArgs

OneFingerCodingWarrior:


OneFingerCodingWarrior:
hi  alan , please check my modification. during my field testing I found there is tens thousands  SocketAsyncEventArgs didn't get GC with vs2017 diagnostic tools. both for debug and release build.

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -1,148 +1,175 @@
...
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                if (null != ReceiveArgs)
+                {
+                    ReceiveArgs.Completed -= HandleOperationCompleted;
+                    SendArgs.Completed -= HandleOperationCompleted;
```
alanmcgovern: If there's a pending Send or Receive then the TaskCompletionSource will never be marked as complete, which means if there's any code calling `await SendAsync` or `await ReceiveAsync` it will hang forever.

If the event handlers are left attached then any pending send/receive call will be aborted automatically, and the event handler will be invoked showing the operation has been cancelled. Similarly, if a send/receive has just completed then it will still invoke the handler and result in the TaskCompletionSource being marked as 'completed'.

I'll add a test showing what i mean...

alanmcgovern: I popped the test and suggestions here: https://github.com/mono/monotorrent/pull/101

OneFingerCodingWarrior: I will re check the memory leaking issue,

alanmcgovern:
I made some small tweaks and merged your suggestion. Thanks for it, it was a good find!

alanmcgovern:
Screenshot 2019-08-01 at 10 51 24

Screenshot 2019-08-01 at 10 55 27

Memory seems stable after over 3 minutes. There are 301 SocketAsyncEventArg objects alive, which seems correct as there were 150 active connections and each active connection has two.

Good stuff!

PR #101, opened 2019-07-31
Dispose with test

alanmcgovern:
@OneFingerCodingWarrior this is a test which shows that 'Send' or 'Receive' hang forever with these changes!

Manually disposing the async args makes sense as an optimisation. If we don't manually dispose then the will be finalised at some point in the future, and we shouldn't wait for that when it's easy to dispose!

I'll put some suggested changes in the commit, if you could make them then we should have a green test and a clean commit! :D

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
@@ -43,19 +43,19 @@ namespace MonoTorrent.Client.Connections
 
         public bool CanReconnect => !IsIncoming;
 
-        public bool Connected => Socket.Connected;
+        public bool Connected => SK.Connected;
 
         EndPoint IConnection.EndPoint => EndPoint;
 
-        public IPEndPoint EndPoint { get; }
+        public IPEndPoint EndPoint { get; private set; }
 
         public bool IsIncoming { get; }
 
-        SocketAsyncEventArgs ReceiveArgs { get; }
+        protected SocketAsyncEventArgs ReceiveArgs { get; private set; }
 
-        SocketAsyncEventArgs SendArgs { get; }
+        protected SocketAsyncEventArgs SendArgs { get; private set; }
 
-        Socket Socket { get; }
+        protected Socket SK { get; private set; }
 
         public Uri Uri { get; protected set; }
 
@@ -65,7 +65,7 @@ namespace MonoTorrent.Client.Connections
 		#region Constructors
 
 		protected SocketConnection(Uri uri)
-            : this (new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp),
+            : this (new Socket((uri.Scheme == "ipv4") ? AddressFamily.InterNetwork : AddressFamily.InterNetworkV6, SocketType.Stream, ProtocolType.Tcp),
                   new IPEndPoint(IPAddress.Parse(uri.Host), uri.Port), false)
 
         {
@@ -88,11 +88,12 @@ namespace MonoTorrent.Client.Connections
             };
             ReceiveArgs.Completed += HandleOperationCompleted;
             SendArgs.Completed += HandleOperationCompleted;
-            Socket = socket;
+            SK = socket;
             EndPoint = endpoint;
             IsIncoming = isIncoming;
         }
 
+        //去掉static函数，否则导致大量的ReceiveArgs以及SendArgs不能释放--因为Dispose函数里也不能释放
         static void HandleOperationCompleted (object sender, SocketAsyncEventArgs e)
         {
             if (e.SocketError != SocketError.Success)
@@ -111,7 +112,7 @@ namespace MonoTorrent.Client.Connections
             var tcs = new TaskCompletionSource<int>();
             ReceiveArgs.UserToken = tcs;
 
-            if (!Socket.ConnectAsync(ReceiveArgs))
+            if (!SK.ConnectAsync(ReceiveArgs))
                 return Task.FromResult(true);
             return tcs.Task;
         }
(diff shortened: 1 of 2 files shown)


src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -43,19 +43,19 @@ public class SocketConnection : IConnection
 
         public bool CanReconnect => !IsIncoming;
 
-        public bool Connected => Socket.Connected;
+        public bool Connected => SK.Connected;
```
alanmcgovern: Can we leave it as `Socket` instead of renaming to `SK`?

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -43,19 +43,19 @@ public class SocketConnection : IConnection
...
         public bool IsIncoming { get; }
 
-        SocketAsyncEventArgs ReceiveArgs { get; }
+        protected SocketAsyncEventArgs ReceiveArgs { get; private set; }
 
-        SocketAsyncEventArgs SendArgs { get; }
+        protected SocketAsyncEventArgs SendArgs { get; private set; }
 
-        Socket Socket { get; }
+        protected Socket SK { get; private set; }
```
alanmcgovern: Do we need to add the `protected` modifier to these? I'd rather leave them private unless there's a really important reason why they have to be public. People are free to implement `IConnection` themselves if they need to do weird things with the underlying sockets. Also, if the SocketAsyncEventArgs are made public then it becomes harder to rewrite the internals of this class to use a different underlying async mechanism in future!

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -65,7 +65,7 @@ public class SocketConnection : IConnection
 		#region Constructors
 
 		protected SocketConnection(Uri uri)
-            : this (new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp),
+            : this (new Socket((uri.Scheme == "ipv4") ? AddressFamily.InterNetwork : AddressFamily.InterNetworkV6, SocketType.Stream, ProtocolType.Tcp),
```
alanmcgovern: I like this! Good catch! Could we make this change in a separate commit?

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -88,11 +88,12 @@ protected SocketConnection(Socket socket, bool isIncoming)
             };
             ReceiveArgs.Completed += HandleOperationCompleted;
             SendArgs.Completed += HandleOperationCompleted;
-            Socket = socket;
+            SK = socket;
             EndPoint = endpoint;
             IsIncoming = isIncoming;
         }
 
+        //去掉static函数，否则导致大量的ReceiveArgs以及SendArgs不能释放--因为Dispose函数里也不能释放
```
alanmcgovern: I'd probably remove this from the PR because i don't understand it :p

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -133,14 +134,40 @@ public Task<int> SendAsync(byte[] buffer, int offset, int count)
...
+                    SendArgs.RemoteEndPoint = null;
+
+                    ReceiveArgs.Dispose();
+                    SendArgs.Dispose();
+
+                    ReceiveArgs = null;
+                    SendArgs = null;
+
+                    //EndPoint = null;//reserved for Logging
+                    Uri = null;
```
alanmcgovern: There's probably no need to null this out

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -133,14 +134,40 @@ public Task<int> SendAsync(byte[] buffer, int offset, int count)
...
-            Socket.Dispose ();
+            try
+            {
+                if (null != ReceiveArgs)
+                {
+                    ReceiveArgs.Completed -= HandleOperationCompleted;
+                    SendArgs.Completed -= HandleOperationCompleted;
+
+                    ReceiveArgs.RemoteEndPoint = null;
+                    SendArgs.RemoteEndPoint = null;
```
alanmcgovern: There should be no need to null this out either

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
```diff
@@ -133,14 +134,40 @@ public Task<int> SendAsync(byte[] buffer, int offset, int count)
...
 
         public void Dispose()
         {
-            Socket.Dispose ();
+            try
+            {
+                if (null != ReceiveArgs)
+                {
+                    ReceiveArgs.Completed -= HandleOperationCompleted;
+                    SendArgs.Completed -= HandleOperationCompleted;
```
alanmcgovern: If we remove the `Completed -=` lines then the test should pass, and everything should be leak-free.

As `HandleOperationCompleted` is a static function then no additional memory would be accidentally retained if we *do not* remove the event handler. Also, if we *do not* remove it then we can be safely guaranteed that it will be invoked exactly once for every `SendAsync` or `ReceiveAsync` call. This means the `TaskCompletionSource` will always have a result and be marked as complete.

OneFingerCodingWarrior:
I forgotten one modification
ClientEngine.cs
//ConnectionManager.CancelPendingConnects();

I do the changes ,  because the process memory occupied raised from 20mb to 40mb and more, after attaching vs2017 diagnostic tools, I found tens of thousands SocketAsyncEventArgs 

in fact , after the modification, the issue gone

src/MonoTorrent/MonoTorrent.Client/PeerConnections/SocketConnection.cs
OneFingerCodingWarrior: purely for I can set SK = null;

alanmcgovern:
Hrmm, interesting. It might be worth it for me to take a look too. I'm curious to see what's happening there!

alanmcgovern:
Actually, I went ahead and made the changes as they're quite minor! I thought it'd be easier :) I left you as the author of the commit and just amended the message to show I made changes too.

PR #103, opened 2019-08-02
Issue 102 fix leaks

alanmcgovern:
@OneFingerCodingWarrior There was one leak where PeerIds would be retained forever in the StandardPicker when we swapped from `StandardPicker` -> `EndGamePicker`. There was a missing call to reset the StandardPicker when the swap occurred :)

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/PiecePicking/EndGamePicker.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/PiecePicking/EndGamePicker.cs
@@ -28,6 +28,8 @@
 
 
 using System;
+using System.Linq;
+
 using System.Collections.Generic;
 using System.Text;
 using MonoTorrent.Client.Messages;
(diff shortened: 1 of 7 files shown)


src/MonoTorrent/MonoTorrent.Client/PiecePicking/EndGamePicker.cs
```diff
@@ -113,7 +113,7 @@ public override void Initialise(BitField bitfield, TorrentFile[] files, IEnumera
             foreach (Piece piece in pieces)
             {
                 for (int i = 0; i < piece.BlockCount; i++)
-                    if (piece.Blocks[i].RequestedOff != null)
+                    if (piece.Blocks[i].RequestedOff != null && !piece.Blocks[i].Received)
```
alanmcgovern: @OneFingerCodingWarrior This was one of the problems i believe!

alanmcgovern:
I ran some profiling checks on the latest revision of code and things look like they're being garbage collected as expected now. `PeerId` objects are going away, and there are no `HaveMessages` being retained in memory either.

While downloading:
Screenshot 2019-08-02 at 14 20 29

A few seconds after downloading completes:
Screenshot 2019-08-02 at 14 21 12

If you can reproduce issues where objects are being retained in memory, if you share a copy of the
retention graph for the object (the part which shows you where/why the objects are not being garbage collected) i might be able to figure out what's going wrong. For example, when downloading completes there are still 40+ BitFields alive. If i look at the retention chart I can see they're all in either the TorrentManager or the PiecePicker, which is what I would expect!

Screenshot 2019-08-02 at 14 25 00

PR #119, opened 2019-08-24
fixed problem in AddPeerAsync with peers from private trackers

### LINKED ISSUE #118: Private torrent
So after successfully downloading a public ubuntu distro torrent, I decided to test out the private torrent that our team is going to use, and I encountered some problem.

`System.InvalidOperationException` with message `You cannot add external peers to a private torrent` was being thrown for each peer. 
On line 296 inside the `Initialize` function in the `TorrentManager.cs`

Our current setup:

(Everything is being done in one local network, for simplicity)
1 raspberry pi 3 b+ running our own little private tracker(opentracker)
1 raspberry pi 3 b+ running Transmission torrent client
1 Chromebook running Transmission torrent client
1 Windows machine running Transmission torrent client

I create a private torrent with our tracker address added in the tracker list from the Transmission client, load that file on all computers. 
3 Transmission clients on 3 different OSes will happily share among themselves.(We usually start seeding from windows client, but that doesn't really matter i guess)

Now I load the same torrent file onto the `SampleClient` it gives me the above error and shuts down.

What I tried so far was to disable DHT and local peer discovery in the ClientEngine.cs file with the boolean flags. 
And I also tried commenting out the portions of the code that loaded the DHT file.

One strange thing is that whenever the error occurs I can see the list of peers and their IPs and they all match the 3 clients we are running for testing. which confuses me even further.

I think monotorrent does support the private torrent and private trackers, or am i wrong?

skelder:
Fixes #118 

Now it should not have issues with adding peers from private trackers.

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
@@ -612,10 +612,13 @@ namespace MonoTorrent.Client
 
         #region Internal Methods
 
-        public async Task<bool> AddPeerAsync (Peer peer)
+        public Task<bool> AddPeerAsync(Peer peer)
+            => AddPeerAsync(peer, false);
+
+        internal async Task<bool> AddPeerAsync (Peer peer, bool fromTrackers)
         {
             Check.Peer (peer);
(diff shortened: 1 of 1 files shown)


alanmcgovern:
Awesome, thanks!

PR #120, opened 2019-08-25
Fix private torrent peers

alanmcgovern:


src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
```diff
@@ -304,7 +304,7 @@ void Initialise(string savePath, string baseDirectory, IList<RawTrackerTier> ann
                     await ClientEngine.MainLoop;
 
                     Peers.BusyPeers.Clear ();
-                    int count = await AddPeersAsync (e.Peers);
+                    int count = await AddPeersAsync (e.Peers, true);
```
alanmcgovern: @skelder I think you left this out of your previous PR, so I added it and also added tests covering the various ways of adding peers :)

PR #132, opened 2019-09-05
Update tracker api

alanmcgovern:
@justscribe These are the changes I suggested in the other PR. This is an API break, but i think it's one of the final API updates I will be making since the old 0.9 release.


justscribe:
Looks good. No problem to change a little my project)

PR #134, opened 2019-09-06
Use a random seed for generating peer id

borigas:
1st PR here. If there's something you'd like done differently, let me know.

The same peer id is always generated (`-MO1000-676515838224`) because `Random` is always seeded with 0 for the 1st `ClientEngine`. This prevents a monotorrent client from being able to connect to other monotorrent clients. Instead, set the 1st random seed randomly and increment from there for addition ClientEngines


alanmcgovern:
Oh good call. This is an embarrassing one :P

alanmcgovern:
Tagged and released :)

PR #135, opened 2019-09-11
Restore old missed but needed functionality

catholic:


src/MonoTorrent/MonoTorrent.Client/Managers/PieceManager.cs
```diff
@@ -181,5 +181,31 @@ public async Task<int> CurrentRequestCountAsync()
             await ClientEngine.MainLoop;
             return Picker.CurrentRequestCount();
         }
+        
+        public T GetPicker<T>() where T : PiecePicker
+        {
+            if (CheckType(Picker, out T result))
+                return result;
+
+            var p = Picker;
```
alanmcgovern: This API would cause issues as it is normal to have multiple pickers of the same type. 

https://github.com/mono/monotorrent/blob/0558daa145ebf16863e438e7fcc6cf5a3ec6ddbf/src/MonoTorrent/MonoTorrent.Client/Managers/PieceManager.cs#L166

What are you trying to accomplish and can we support this too?

catholic: I need to change some properties of current PiecePicker dynamically outside of picker class.

If this change can cause problems, changes to PiecePicker.cs and PieceManager.cs can be omitted.
I'm able to accomplish this task also using TorrentManager.PieceManager.Picker with some GetPicker() method modifications and using it outside of PieceManager class

alanmcgovern: That makes sense. Out of interest, which properties are you trying to modify dynamically? Are they on the built-in types or are you creating a custom picker?

My original thinking was that the built in pickers don't really have properties that need dynamic changing, but if the use implemented their own picker they'd store it in a field/something before passing it to the PieceManager. As such they wouldn't really need to look it up later. 

Im open to any changes to make things easier :)

catholic: I'm using my own custom picker. It's a combination of SlidingWindow and EndGame pickers in mind, but uses more complex logic.
My picker has some public properties which are modified many times during the file download process. Circumstances of what pieces to download and when are accumulated in another service which modifies Picker properties.

Another approach is to use EventHandler in my service and subscribe on it inside Picker, but I don't like this idea.

alanmcgovern:
LGTM! I'm just travelling at the moment so i was slow to respond after your update.

The additional changes look great, I'll merge them in now!

PR #147, opened 2019-10-12
[core] Add a PartialProgress public property

alanmcgovern:
When some files are marked as DoNotDownload then this property
tracks the completion rate for the files which are marked as
downloadable.

If all downloadable files have been downloaded, the state will
change to 'Seeding' without telling the tracker the download has
completed.

If files which used to be marked as 'DoNotDownload' are then marked
as downloadable, the state will change back to Downloading.

### DIFF
--- a/src/MonoTorrent.Tests/MonoTorrent.Client.Modes/DownloadModeTests.cs
+++ b/src/MonoTorrent.Tests/MonoTorrent.Client.Modes/DownloadModeTests.cs
@@ -28,6 +28,7 @@
 
 
 using System;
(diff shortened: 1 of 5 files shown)


alanmcgovern:
@justscribe This adds and documents the `PartialProgress` property which should give you the information you need. `TorrentManager.State` will change between Seeding and Downloading based on the `PartialProgress` too.

alanmcgovern:
I'll assume this has no issues and if there are any further changes necessary we'll do them in another PR.

PR #148, opened 2019-10-25
Vastly reduce allocations from async/await infrastructure

alanmcgovern:
Setup:
* Debug build
* Only peers on my local network are used, no internet based peers.
* Two 2GB torrents fully downloaded

Memory allocations before changes:
Memory_Unoptimised

Memory allocations after changes:
Memory_Optimised

### DIFF
--- a/src/MonoTorrent.Tests/Client/DiskManagerExceptionTests.cs
+++ b/src/MonoTorrent.Tests/Client/DiskManagerExceptionTests.cs
@@ -163,14 +163,14 @@ namespace MonoTorrent.Client
         public void ReadFail()
         {
             writer.read = true;
-            Assert.ThrowsAsync<Exception> (() => diskManager.ReadAsync (data, 0, buffer, buffer.Length));
+            Assert.ThrowsAsync<Exception> (() => diskManager.ReadAsync (data, 0, buffer, buffer.Length).AsTask ());
         }
 
         [Test]
         public void WriteFail()
         {
             writer.write = true;
-            Assert.ThrowsAsync<Exception> (() => diskManager.WriteAsync(data, 0, buffer, buffer.Length));
+            Assert.ThrowsAsync<Exception> (() => diskManager.WriteAsync(data, 0, buffer, buffer.Length).AsTask ());
         }
     }
 }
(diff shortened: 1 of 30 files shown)


alanmcgovern:
This isn't 100% comparable to the previous screenshots. The build used for this profiling attempt was a Release build (the previous were Debug builds) and I downloaded the torrent normally rather than restricting it to downloading from clients on my local network.

The results are better even though this should be a *worse* setup than the original testing :)

Memory

Therzok:
@alanmcgovern have you tried looking at the performance numbers on release builds _before_ this commit?

Therzok:
There's a lot of overhead in the async debug machinery, wondering what the differences in perf characteristics are.

alanmcgovern:
@Therzok Done! I updated the previous comment so that both comments have like-with-like comparisons for before/after.

alanmcgovern:
I'm debating pushing this out as it's own NuGet... because why not!

This was a fairly rough prototype that I hammered together without really understanding how all the pieces hung together. There's probably some cleanup which could be done, especially around `SetException` handling in `AsyncVoidMethodBuilder` before this would be fully production ready.

At this point it's a nearly 5x reduction in memory allocations (not memory *usage*!).  The restrictions enforced by `ReusableTask` are pretty much irrelevant to real code as you can call `AsTask ()` when you want to do something that `ReusableTask` itself doesn't allow (such as calling `await` multiple times, or if you want to have a blocking `Wait` or `Result` invocation), though i suspect `ReusableTask.AsTask ()` would still use *less* memory overall than using plain Task's everywhere.

`ReusableTaskCompletionSource<int>` are being garbage collected as two are allocated per `SocketConnection` instance, and when that is GC'ed the TCS's are also GC'ed with no attempt to re-cache them. If you ignore the top 4 items based on that, then you can see very few objects are allocated, and nearly none are freed :) 

Filtering on `ReusableTask`
ReusableTasks

Filtering on `ResultHolder`.
ResultHolder

Therzok:
Yep, all looks really nice. Congrats!

alanmcgovern:
I moved the implementation of `ReusableTask` https://github.com/alanmcgovern/ReusableTasks and released it as a standalone NuGet.

I then rewrote/rebased this PR to use the released NuGet instead of including the source directly.

alanmcgovern:
yolo!

PR #151, opened 2019-10-30
Biginteger

alanmcgovern:
Scenario:
* Release build
* Downloading one 2GB torrent and one 1.2GB torrent.

Using the old BigInteger implementation (on top of nearly all the changes to use `ReusableTask`):
MonoTorrent BigInteger

Using the new BigInteger implementation (on top of all the recent changes to use `ReusableTask`):
System Numerics BigInteger

These were taken with DHT enabled.


alanmcgovern:
YOLO :)

PR #152, opened 2019-10-30
[core] Re-enable webseeding

alanmcgovern:
Addresses https://github.com/mono/monotorrent/issues/149

Screenshot 2019-10-30 at 22 56 28

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
@@ -98,6 +98,16 @@ namespace MonoTorrent.Client
         }
 
         /// <summary>
+        /// The delay before a torrent will start using web seeds.
+        /// </summary>
+        public TimeSpan WebSeedDelay { get; set; } = TimeSpan.FromMinutes (1);
+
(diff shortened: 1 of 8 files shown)


alanmcgovern:
@samhocevar One of the issues is that I don't know of any torrents which have web seeds. The only way I can test is to load up a ubuntu torrent and then hardcode in the url for the equivalent http download for that file. If you have real world samples, just load them up and see if it works!

src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
```diff
@@ -97,6 +97,16 @@ public class TorrentSettings : ICloneable
...
 
+        /// <summary>
+        /// The delay before a torrent will start using web seeds.
+        /// </summary>
+        public TimeSpan WebSeedDelay { get; set; } = TimeSpan.FromMinutes (1);
+
+        /// <summary>
+        /// The download speed under which a torrent will start using web seeds.
+        /// </summary>
+        public int WebSeedSpeedTrigger { get; set; } = 15 * 1024;
```
alanmcgovern: 💯

PR #154, opened 2019-10-31
Torrent stream position

justscribe:
When loading torrent the stream postion must be zero.


justscribe:
Yes, you're convinced me. Thanks.

PR #171, opened 2019-11-11
[core] Move TorrentCreator to read and hash in parallel

alanmcgovern:
This is a v1 refactor to test the idea of putting disk IO in one thread and SHA1 hashing in a second thread when hashing torrents.

The original implementation parallelised on each 16kB chunks as I was investigating what it would take to create merkle trees as per bittorrent v2 spec[0]. However, normal bittorrent hashing can only parallelise on entire pieces, which will require a little bit more work in order to get the same scaling benefits as the merkle tree approach received.

[0] tl;dr i don't see any benefit in implementing this as only one other client has attempted to do so, and the spec is so poor it doesn't provide a usable description of the new messages it has introduced. The only way to implement it is to look at the only pre-existing implementation and replicate the approach/layout they used for the new messages, or guess and probably be incompatible.

### DIFF
--- a/src/MonoTorrent/MonoTorrent/TorrentCreator.cs
+++ b/src/MonoTorrent/MonoTorrent/TorrentCreator.cs
@@ -29,10 +29,10 @@
 
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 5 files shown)


src/MonoTorrent/MonoTorrent/TorrentCreator.cs
```diff
@@ -180,62 +181,108 @@ void AddCommonStuff (BEncodedDictionary torrent)
...
+            int parallelTasks = Environment.ProcessorCount;
+            var emptyBuffers = new AsyncProducerConsumerQueue<byte[]> (parallelTasks + 1);
+            var filledBuffers = new AsyncProducerConsumerQueue<(byte[], int, TorrentFile)> (parallelTasks);
 
-            try {
-                foreach (TorrentFile file in files) {
-                    fileRead = 0;
-                    md5Hasher?.Initialize ();
+            for (int i = 0; i < parallelTasks; i ++)
+                await emptyBuffers.EnqueueAsync (new byte[BlockSize]);
```
Therzok: You could use ArrayPool here.

alanmcgovern: Yup, it's definitely a good candidate for buffer re-use, and also a good candidate for `ArrayPool`.

`ArrayPool` isn't built into .NET Standard 2.0, though it is available as a NuGet. I'm a little dependency averse so i've been resisting the urge to delete [pre-existing implementations](https://github.com/mono/monotorrent/blob/master/src/MonoTorrent/MonoTorrent.Client/Managers/BufferManager.cs) and replace them with NuGets.

Apart from that, I decided to not use pooled buffers while I work the underlying changes. It should be relatively easy to swap to pooled buffers when this is done :)

alanmcgovern:
Closing this in favour of https://github.com/mono/monotorrent/pull/177. It's basically the same, but done better.

PR #172, opened 2019-11-12
[core] Do not ignore FileNotFoundException when discovering local peers.

samhocevar:
The previous catch-all exception handling would enter an infinite loop when
for some reason ReusableTasks.dll was not properly deployed.

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/LocalPeerDiscovery.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/LocalPeerDiscovery.cs
@@ -153,8 +153,8 @@ namespace MonoTorrent.Client
                     var uri = new Uri("ipv4://" + result.RemoteEndPoint.Address + ':' + portcheck);
 
                     PeerFound?.InvokeAsync (this, new LocalPeerFoundEventArgs (infoHash, uri));
-                } catch (FileNotFoundException ex) {
-                    throw ex;
+                } catch (FileNotFoundException) {
+                    throw;
                 } catch {
 
                 }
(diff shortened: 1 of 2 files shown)


samhocevar:
I know this is not a very elegant fix. But since `ReusableTasks.dll` is only loaded on-demand by the CLR, an incomplete deployment can be detected quite late in the program’s lifetime.

Another suggestion would be to add a dummy public variable such as a version number in `ReusableTasks.dll` that gets imported from `MonoTorrent.TaskExtensions`, but that’s not ideal either.

alanmcgovern:
I do like the idea of ensuring this doesn't end up as an infinite loop. I should be tighter on the types of exceptions. 

In addition to this, one way to make this occur earlier would be to use a ReusableTask somewhere internally in the ClientEngine class. 

If the ClientEngine.Register, or StartAll, or the constructor, called GC.KeepAlive (ReusableTask.CompletedTask) then we could guarantee you'll get a correctly propagated exception before you get too far into the codebase. I'll take your patch, add that change and then we should be good to go with the sanity-checks! I won't get around to that until tomorrow though :)

src/MonoTorrent/MonoTorrent.Client/LocalPeerDiscovery.cs
```diff
@@ -152,6 +153,8 @@ async void ReceiveAsync (UdpClient client, CancellationToken token)
                     var uri = new Uri("ipv4://" + result.RemoteEndPoint.Address + ':' + portcheck);
 
                     PeerFound?.InvokeAsync (this, new LocalPeerFoundEventArgs (infoHash, uri));
+                } catch (FileNotFoundException ex) {
+                    throw ex;
```
alanmcgovern: This should be `throw` so the full stacktrace is preserved, instead of `throw ex` which will lose the original stacktrace :) I'll tweak it and commit the changes!

PR #192, opened 2019-12-02
Add "forced" parameter for TorrentManager.StopAsync() and StoppingMod…

catholic:
…e.WaitForStoppingToComplete() methods to speed-up torrent stop process

Sometimes announcing takes too long. It will be useful to give ability in MonoTorrent API to force the stop process by not sending announce to Tracker.

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
@@ -637,7 +637,8 @@ namespace MonoTorrent.Client
         /// <summary>
         /// Stops the TorrentManager. The returned task completes as soon as the manager has fully stopped.
         /// </summary>
-        public async Task StopAsync()
+        /// <param name="forced">If set as true, Tracker will not be announced with Stopped torrent event</param>
+        public async Task StopAsync(bool forced = false)
         {
             await ClientEngine.MainLoop;
 
@@ -650,7 +651,7 @@ namespace MonoTorrent.Client
             } else if (State != TorrentState.Stopped) {
                 var stoppingMode = new StoppingMode(this, Engine.DiskManager, Engine.ConnectionManager, Engine.Settings);
                 Mode = stoppingMode;
-                await stoppingMode.WaitForStoppingToComplete ();
+                await stoppingMode.WaitForStoppingToComplete(forced);
 
                 Mode = new StoppedMode (this, Engine.DiskManager, Engine.ConnectionManager, Engine.Settings);
                 Engine.Stop();
--- a/src/MonoTorrent/MonoTorrent.Client/Modes/StoppingMode.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Modes/StoppingMode.cs
@@ -47,27 +47,42 @@ namespace MonoTorrent.Client.Modes
             CanAcceptConnections = false;
         }
 
-        public async Task WaitForStoppingToComplete ()
+        /// <summary>
+        /// </summary>
+        /// <param name="forced">If set as true, Tracker will not be announced with Stopped torrent event</param>
+        public async Task WaitForStoppingToComplete(bool forced = false)
         {
             try {
-                Manager.Engine.ConnectionManager.CancelPendingConnects (Manager);
-                foreach (PeerId id in Manager.Peers.ConnectedPeers.ToArray ())
-                    Manager.Engine.ConnectionManager.CleanupSocket (Manager, id);
+                Manager.Engine.ConnectionManager.CancelPendingConnects(Manager);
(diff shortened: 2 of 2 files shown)


alanmcgovern:
I don't think it's a good idea to build in the ability to unilaterally disable sending the Stopped event.

There's no clear logic for a user of the MonoTorrent library to easily figure out when to pass `true` or when to pass `false`, which means the most likely outcome will be that a developer will hardcode `true` at some point and then their application will end up banned by trackers as it does not send the correct announce requests when stopping.

At the moment the maximum possible time you should need to wait, no matter how many torrents you have, is [60 seconds](https://github.com/mono/monotorrent/blob/master/src/MonoTorrent/MonoTorrent.Client/Modes/StoppingMode.cs#L67). You can do this invoking [ClientEngine.StopAll](https://github.com/mono/monotorrent/blob/master/src/MonoTorrent/MonoTorrent.Client/ClientEngine.cs#L409-L418), or just replicate the logic of calling `StopAsync` on every TorrentManager and then awaiting all the tasks in one go.

If you do something like:
```
foreach (TorrentManager manager in managers)
    await manager.StopAsync ();
```
then the maximum time this will take will be 60 seconds per manager, which would be pretty bad if you have a lot of torrents.

If you are calling `ClientEngine.StopAll`, or the equivalent, and want it to be faster than 1 minute, then I could make this value configurable and enforce a reasonable minimum time of 15 seconds (or so) to ensure that under normal circumstances the final announce should be correctly processed.

I can probably enhance things so that the final 'Stopped' announce is skipped completely if no tracker has been successfully announced to before.

catholic:
For my needs stop process should take not more than 3 seconds. 

> want it to be faster than 1 minute, then I could make this value configurable

I think this could be a solution.

>I can probably enhance things so that the final 'Stopped' announce is skipped completely if no tracker has been successfully announced to before.

I suppose it will be unneeded overhead in the stop process.

alanmcgovern:
3 seconds may not be long enough for busy trackers which are online and active. Why do you need the time to be so short? Is there a restriction you're trying to meet?

catholic:
>Why do you need the time to be so short?

I need to give my users experience of torrent files usage as of media files. They can frequently start and stop them, add and remove from playlist. Adding torrent to a playlist also Registers TorrentManager in ClientEngine. Removing torrent from playlist also Unregisters TorrentManager from ClientEngine.

alanmcgovern:
Just wanted to drop a comment on this and say I haven't forgotten about this. I just need to find the time to complete the changes :)

alanmcgovern:
Hi! Thanks for bearing with me as I looked into this! I've made two sets of changes.

1) https://github.com/mono/monotorrent/pull/211

This ensures that we keep the tracker status updated as best we can. If our most recent scrape/announce indicates the tracker is offline, then the final announce will be skipped as per the pre-existing logic. The real change here is that UdpTracker and HttpTracker have been updated to more accurately track the state, so we should hit that fastpath more frequently (especially for UdpTracker!).

2) https://github.com/mono/monotorrent/pull/214
This gives you the timeout you needed. The *minimum* enforced by the library is currently 2 seconds. I reserve the right to arbitrarily change this in future ;)  However, if you wish to enforce a timeout during the 'Stop' process you can now call `ClientEngine.StopAllAsync(TimeSpan)` or `TorrentManager.StopAsync (TimeSpan)`. The default is `Timeout.InfiniteTimeSpan` but any positive value is allowed, and anything lower than 2 seconds will currently be clamped to 2 seconds.

```
I need to give my users experience of torrent files usage as of media files. They can frequently start and stop them, add and remove from playlist. Adding torrent to a playlist also Registers TorrentManager in ClientEngine. Removing torrent from playlist also Unregisters TorrentManager from ClientEngine.
```
One thing you could try would be to filter out any `TorrentManager` instances from your UI if they have a state of `Stopping`. This would let you pretend the actions are happening faster than the really are.

If the combination of these two PRs is good enough to meet your needs then I'd like close this PR and not merge it. I worry that if the final announce can be skipped by setting a boolean, then someone would leave it set to `true` and then their client could be banned from trackers due to it never sending the final announce.

catholic:
> If the combination of these two PRs is good enough to meet your needs then I'd like close this PR and not merge it.

Yes, this seems very good and enough for my needs. Thank you!

> One thing you could try would be to filter out any TorrentManager instances from your UI if they have a state of Stopping. This would let you pretend the actions are happening faster than the really are.

This option was not good previously, because a user could remove TorrentManager and seconds after try to add it again while TorrentManager still in Stopping mode. But with a combination of 2 seconds of Timeout, this can be done smoothly now.

alanmcgovern:
Brilliant! I'll close this now and if you still need some tweaks do let me know :)

PR #200, opened 2019-12-16
[core] Don't special case peers which are handshaking

alanmcgovern:
Once a connection has been established to a remote peer, or
an incoming connection has been received, we should add the
peer to the 'Connected' list.

This way any logic which attempts to close all open connections
will successfully close connections which are still completing
the normal bittorrent handshake.


alanmcgovern:
More thorough fix for issues like https://github.com/mono/monotorrent/issues/196 .

I don't know why the 'handshakingpeers' list was added in the first place. It showed up as part of a refactor with no explanation :p

PR #204, opened 2019-12-28
Fixed: Prevent exception if nodes is empty string

Qstick:
Reference https://github.com/Sonarr/Sonarr/commit/3bc6bf9e992b0c2e0f82591608f2037fdd313297

### DIFF
--- a/src/MonoTorrent/MonoTorrent/Torrent.cs
+++ b/src/MonoTorrent/MonoTorrent/Torrent.cs
@@ -801,7 +801,8 @@ namespace MonoTorrent
                             break;
 
                         case ("nodes"):
-                            nodes = (BEncodedList)keypair.Value;
+                            if (keypair.Value.ToString().Length != 0)
+                                this.nodes = (BEncodedList)keypair.Value;
                             break;
 
                         case ("comment.utf-8"):


alanmcgovern:
@Qstick My only question is why is this a problem in the first place? If the data contains an empty `BEncodedList` then is some other part of the code hitting an `IndexOutOfRangeException` by assuming it's non-empty, or is the issue that some application is (incorrectly?) putting in a `BEncodedString` instead of a `BEnodedList` and you're getting an `InvalidCastException` ?

alanmcgovern:
Possibly fixed by https://github.com/mono/monotorrent/commit/b1f976a18fd42bef1c9c54a6bae654a3221283f6 . If not, let me know what the issue is :)

alanmcgovern:
This should be fixed with the patch in master. I was linked to this as the probable original issue ( https://forums.sonarr.tv/t/sonarr-fails-to-download-torrent-files-from-rss-feed/9700 ) and it indicates an InvalidCastException was the cause. That makes me believe the issue is that some implementors of this feature use a BEncodedList of BEncodedLists and others used a BEncodedString with compact encoding. The spec doesn't really say what you should do, but it heavily implies BEncodedList of BEncodedLists.

If there are problematic torrents do let me know and we can figure out the next step! For now i think things should be OK.

alanmcgovern:
Thanks for the patch, even though I didn't take it exactly as-is this time :)

PR #208, opened 2019-12-29
Fixes mono/monotorrent#207

jpmikkers:


alanmcgovern:
Thanks! Good catch!

alanmcgovern:
I added tests in https://github.com/mono/monotorrent/commit/4f7383eff88546832d79d0989d996f78214df87e too. Now pausing is covered for the hashing, downloading and seeding states.

PR #213, opened 2020-01-15
[core] Add path validation when loading torrents

alanmcgovern:
Ensure path traversal attacks are prevented by disallowing
torrents with embedded path traversals. The logic is shared
between TorrentCreator and Torrent so the same validations
are performed when loading or creating torrents.


alanmcgovern:
Addresses https://github.com/mono/monotorrent/issues/212

PR #214, opened 2020-01-16
[core] Add a timeout when stopping torrents

alanmcgovern:
This allows a relatively simple override for the final announce
so that the returned task doesn't hang around forever.

This is not implemented in the best of ways. Ideally the
ITracker and ITrackerManager interfaces would be augmented
to take a CancellationToken or TimeSpan parameter to handle
cancellation/timeouts, but that would be an ABI break which
I don't want to do right now.

This approach isn't great as it doesn't actually cancel the
pending request, but it's good enough and allows people to
use the timeout functionality in a way which will be
compatible with any future changes to ITracker/ITrackerManager.

### DIFF
--- a/src/MonoTorrent.Tests/Client/ManualTrackerManager.cs
+++ b/src/MonoTorrent.Tests/Client/ManualTrackerManager.cs
@@ -49,6 +49,8 @@ namespace MonoTorrent.Client
 
         public DateTime LastUpdated { get; }
 
+        public TimeSpan ResponseDelay { get; set; }
+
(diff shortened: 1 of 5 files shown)


alanmcgovern:
@catholic Will this work for you instead of the `bool` parameter which always skips the final announce?

catholic:
> Will this work for you instead of the bool parameter which always skips the final announce?

Yes. Thank you!

alanmcgovern:
Brilliant! I'll get this released in a nuget soon :)

PR #220, opened 2020-01-23
Add a way to stream media while downloading

alanmcgovern:
This is a prototype which hasn't even been tested yet ;)

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
+++ b/src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
@@ -1,4 +1,33 @@
-﻿using System;
+﻿//
+// LocalStream.cs
+//
+// Authors:
+//   Alan McGovern alan.mcgovern@gmail.com
+//
+// Copyright (C) 2020 Alan McGovern
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+
+using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


alanmcgovern:
@uponatime2019 I think this could help you out!

Firstly - this is currently *completed* untested :p Don't worry if it's completely broken. I'm letting you know early as it shows you what I intend to offer!

The idea is that you can create a `ClientEngine` and `TorrentManager` as per normal, then you do something like:

```
public async Task<Stream> CreateStreamAsync (TorrentManager manager)
{
    var file = manager.Torrent.Files [0]; // Choose the file you want to stream
    var stream = new DirectStream (manager, file);
    await stream.Initialise ();
    return stream;
}
```

Then you can pass that stream directly to UWP (or whatever). This stream supports seeking, and will ensure the data has been downloaded before the `Read` method returns. I'd love for you to give this a shot once i complete the implementation and testing.

If you want to stream multiple different files from one torrentmanager you'll have to dispose the first `DirectStream` before you create a second one. The `TorrentManager` will not be disposed autmatically, nor will it be stopped, so it should be possible to simply create a second stream and keep going. That code will need reviewing though.

uponatime2019:
Yeah thank you, I will try it

alanmcgovern:
Ok, i think it's kinda sorted ish!

I've been streaming a video to VLC over HTTP with full seeking more or less working as expected. There are still some tweaks which are needed to improve seeking performance, but it works.

You can also create a `Stream` which can be passed to normal .NET APIs. This is also seekable, but the difference is that if the blocking APIs are used (`Stream.Read` instead of `Stream.ReadAsync`) then there's one limitation for seeking: If you seek to position A and Read is invoked before the data is available, the call will block until the data is downloaded. You must wait for this to unblock before you can successfully seek to a new position in the stream.

Apart from that, there is still a fair bit of testing a tweaking needed. However it would be great if you gave it a shot now to see if it can handle single-file streaming in a way which works for you!

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public override void Flush ()
+            => throw new NotSupportedException ();
+
+        public override int Read (byte[] buffer, int offset, int count)
+            => ReadAsync (buffer, offset, count, CancellationToken.None).GetAwaiter ().GetResult ();
+
+        public override async Task<int> ReadAsync (byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            long torrentOffset = 0;
```
ilyalatt: @alanmcgovern Why not `file.StartPieceIndex * PieceLength + file.StartPieceOffset`?

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
```diff
@@ -0,0 +1,153 @@
...
+            foreach (var file in Manager.Torrent.Files) {
+                if (file == File)
+                    break;
+
+                torrentOffset += file.Length;
+            }
+
+            var startPiece = (int) (torrentOffset + offset + Position) / Manager.Torrent.PieceLength;
```
ilyalatt: @alanmcgovern `torrentOffset + offset + Position` is duplicated. Can be moved to a variable like `startOffset`. https://github.com/ilyalatt/BitTorrentStorage/blob/master/src/BitTorrentStorage/BtsFileStream.cs#L92

uponatime2019:
I'm tested with LocalStream but downloader seem not start from first piece (it notice a random piece Piece Index: 1566/524288) and media player don't start. Same result with HttpStream
```
        public static async Task<Stream> CreateLocalStreamAsync(TorrentManager manager, TorrentFile file)
        {
            var provider = new StreamProvider(manager);
            return await provider.CreateStreamAsync(file);
        }
```

And you're using Console.WriteLine to write some some output, which not support on UWP, change you change it to Debug.WriteLine
```
<OUTPUT>
Unsupported standard style: UW
Unsupported standard style: UW
Unsupported standard style: UW
```

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
alanmcgovern: :+1:

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
alanmcgovern: 👍

alanmcgovern:
The output was mostly just for my own benefit when implementing, it wasn't intended to be permanent output. However it probably makes sense to move to Debug.WriteLine for things like this and keep it  in there permanently :)

Apart from that, what kind of media file are you trying to play and are you using the built in Uwp MediaElement API to play it? MonoTorrent itself doesn't do any seeking, so if your media player chooses to seek to the mid-point of a file then (presumably) that's the intention. Does the torrent have multiple files such that the seek corresponds to going to the start of one of the files? Can you share a testcase to help debug what's going on?

alanmcgovern:
feck. Merged it prematurely. Fixing that now! :p

alanmcgovern:
PR re-opened here: https://github.com/alanmcgovern/monotorrent/pull/224

PR #222, opened 2020-01-25
Few fixes and improvements

vlasenkoalexey:
I'm using monotorrent in my App, here are some changes I have made to the latest version.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,12 @@
+# These are supported funding model platforms
+
+github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: # Replace with a single Patreon username
+open_collective: monotorrent
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+otechie: # Replace with a single Otechie username
+custom: # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']


alanmcgovern:
Hi! Some of these changes look quite useful. I'd love to include them and there's one thing you could do to make it easier for me to do so!

In general if one pull request is created per feature/changeset it makes it easier to review, iterate and merge changes. This PR has a single commit with many different changesets This means I have a choice of accepting everything or nothing. I'll annotate the PR with thoughts on how things could be split up for submission. 

Thanks!

src/MonoTorrent/MonoTorrent.Client.Connections/SocketConnection.cs
```diff
@@ -198,16 +198,20 @@ public ReusableTask<int> ReceiveAsync (byte[] buffer, int offset, int count)
             args.SetBuffer (offset, count);
             args.UserToken = ReceiveTcs;
 
+            // https://github.com/alanmcgovern/monotorrent/issues/202
+#if DEBUG
             AsyncFlowControl? control = null;
             if (!ExecutionContext.IsFlowSuppressed ())
                 control = ExecutionContext.SuppressFlow ();
-
+#endif
```
alanmcgovern: If this is wrapped in `#if DEBUG` then it will only have an effect when running local debug builds. The NuGet packages from nuget.org are built with the `Release` configuration so this code will still execute. Is this intended?

Also, this changeset would be good as a separate PR.

src/MonoTorrent/MonoTorrent.Client.Messages/MessageBundle.cs
```diff
@@ -85,5 +86,10 @@ public override int Encode (byte[] buffer, int offset)
 
             return CheckWritten (written - offset);
         }
+
+        public override string ToString()
+        {
+            return $"MessageBundle [{String.Join(", ", this.Messages.Select(m => m.ToString()).ToArray())}]";
```
alanmcgovern: Do you want `ToString` to aid debugging? If so you could implement it similar to how BitField does it ( https://github.com/alanmcgovern/monotorrent/blob/master/src/MonoTorrent/MonoTorrent/BitField.cs#L43 )

src/MonoTorrent/MonoTorrent.Client.Messages/PeerMessage.cs
```diff
@@ -107,5 +107,10 @@ internal void Handle (TorrentManager manager, PeerId id)
         {
             manager.Mode.HandleMessage (id, this);
         }
+
+        public override string ToString()
+        {
+            return this.GetType().Name;
```
alanmcgovern: The default ToString implementation should return the typename. https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=netframework-4.8#remarks . That should mean this change is unnecessary?

src/MonoTorrent/MonoTorrent.Client.Tracker/TrackerTier.cs
```diff
@@ -36,7 +36,6 @@ public class TrackerTier
     {
         public IList<ITracker> Trackers { get; }
 
-        internal bool SendingStartedEvent { get; set; }
         internal bool SentStartedEvent { get; set; }
```
alanmcgovern: :+1:

src/MonoTorrent/MonoTorrent.Client.Tracker/UdpTracker.cs
```diff
@@ -60,6 +60,8 @@ protected override async Task<List<Peer>> DoAnnounceAsync (AnnounceParameters pa
                 var message = new AnnounceMessage (DateTime.Now.GetHashCode (), connectionId, parameters);
                 var announce = (AnnounceResponseMessage) await SendAndReceiveAsync (message);
                 MinUpdateInterval = announce.Interval;
+                Complete = announce.Seeders;
+                Incomplete = announce.Leechers;
```
alanmcgovern: This would be good as a separate PR, ideally with a unit test. I can add one once the change is split up if you do not have the time to do so!

src/MonoTorrent/MonoTorrent.Client/ClientEngine.cs
```diff
@@ -77,6 +78,9 @@ public class ClientEngine : IDisposable
         public event EventHandler<StatsUpdateEventArgs> StatsUpdate;
         public event EventHandler<CriticalExceptionEventArgs> CriticalException;
 
+        public event EventHandler<TorrentEventArgs> TorrentRegistered;
+        public event EventHandler<TorrentEventArgs> TorrentUnregistered;
```
alanmcgovern: This looks like a good addition. Could this, and related changes, be added in a separate PR?

src/MonoTorrent/MonoTorrent.Client/ClientEngine.cs
```diff
@@ -468,7 +475,9 @@ private void LogicTick ()
                 uploadLimiter.UpdateChunks (Settings.MaximumUploadSpeed, TotalUploadSpeed);
             }
 
-            ConnectionManager.CancelPendingConnects ();
+            for (int i = 0; i < this.torrents.Count; i++)
+                ConnectionManager.CancelPendingConnects(this.torrents[i]);
```
alanmcgovern: This change looks like it's broken. It should cause pending connects to be closed every time LogicTick is invoked rather than only when the timeout has been reached. This means pending connections will be closed between 0 and 500ms after the connection attempt is initiated. Do you recall what issue you were trying to solve with this change?

src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
```diff
@@ -286,6 +288,48 @@ public class TorrentManager : IDisposable, IEquatable<TorrentManager>
...
         public InfoHash InfoHash { get; }
 
+        /// <summary>
+        /// Torrent tracker announce Urls
+        /// </summary>
+        public IList<RawTrackerTier> AnnounceUrls
+        {
+            get;
+            private set;
+        }
```
alanmcgovern: This shouldn't be exposed at this level. Why did you need access to this here?

src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
```diff
@@ -406,7 +451,7 @@ void CreateRateLimiters ()
 
         #region Public Methods
 
-        internal void ChangePicker (PiecePicker picker)
+        public void ChangePicker(PiecePicker picker)
```
alanmcgovern: This should not be public as it is not threadsafe for users of the library to invoke this directly. A threadsafe method does exist (`ChangePickerAsync`) so that's what people should be using.

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
```diff
@@ -139,6 +139,10 @@ async Task Announce (TorrentEvent clientEvent, ITracker referenceTracker)
                         actualArgs = actualArgs.WithClientEvent (TorrentEvent.Started);
 
                     var peers = await tuple.Item2.AnnounceAsync (actualArgs);
+                    if (clientEvent == TorrentEvent.Started)
+                    {
+                        tuple.Item1.SentStartedEvent = true;
```
alanmcgovern: This looks good! It could be it's own PR too.

src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
```diff
@@ -286,6 +288,48 @@ public class TorrentManager : IDisposable, IEquatable<TorrentManager>
...
+        /// <summary>
+        /// Magnet link
+        /// </summary>
+        public MagnetLink MagnetLink {
+            get {
+                if (this.magnetLink == null) {
+                    this.magnetLink = new MagnetLink (Torrent.InfoHash, name: Name,
+                        announceUrls: AnnounceUrls != null ? AnnounceUrls.SelectMany (rawTrackerTier => rawTrackerTier).ToList () : new List<string> ());
+                }
+                return this.magnetLink;
```
alanmcgovern: This set of changes looks good and would be ideal as it's own PR!

src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
```diff
@@ -286,6 +288,48 @@ public class TorrentManager : IDisposable, IEquatable<TorrentManager>
...
+            private set;
+        }
+
+        /// <summary>
+        /// Magnet link
+        /// </summary>
+        public MagnetLink MagnetLink {
+            get {
+                if (this.magnetLink == null) {
+                    this.magnetLink = new MagnetLink (Torrent.InfoHash, name: Name,
```
alanmcgovern: Couldn't we just invoke `Torrent.ToMagnetLink` rather than reimplementing it? This method does rely on `Torrent` being non-null so it should simpler!

vlasenkoalexey:
sg, I'll will split this PR into separate ones for each specific change one I have some time.

src/MonoTorrent/MonoTorrent.Client.Connections/SocketConnection.cs
alanmcgovern: A similar change has been integrated and released in the latest nuget package. There is a bug in some older versions of UWP which causes nullreference exceptions if you use this API. I assume you encountered that issue :)

src/MonoTorrent/MonoTorrent.Client.Tracker/TrackerTier.cs
alanmcgovern: I removed this variable :)

PR #224, opened 2020-01-26
Streaming

alanmcgovern:
Recreating https://github.com/alanmcgovern/monotorrent/pull/220 after an accidental merge.


src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
```diff
@@ -0,0 +1,155 @@
...
+        public override void Flush ()
+            => throw new NotSupportedException ();
+
+        public override int Read (byte[] buffer, int offset, int count)
+            => ReadAsync (buffer, offset, count, CancellationToken.None).GetAwaiter ().GetResult ();
+
+        public override async Task<int> ReadAsync (byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            // The torrent is treated as one big block of data, so this is the offset at which the current file's data starts at.
+            var torrentFileStartOffset = (long)File.StartPieceIndex * (long)Manager.Torrent.PieceLength + File.StartPieceOffset;
```
ilyalatt: `count` can be more then `Length - Position`. Moreover, there is a code in .net fw (i do not remember where) that passes buffer size as `count` and cycles while the result (bytes read count) is greater then 0. Also `Position` can be greater or equal then `Length`. You can check this things in the documentation, [here](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.position?view=netframework-4.8#remarks) and [here](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=netframework-4.8#System_IO_Stream_Read_System_Byte___System_Int32_System_Int32_). Here is [the example](https://github.com/ilyalatt/BitTorrentStorage/blob/master/src/BitTorrentStorage/BtsFileStream.cs#L111).

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
```diff
@@ -0,0 +1,155 @@
...
+
+                if (allAvailable)
+                    break;
+
+                await Task.Delay (500, cancellationToken);
+            }
+
+            cancellationToken.ThrowIfCancellationRequested ();
+            if (Stream == null) {
+                Stream = new FileStream (File.FullPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
```
ilyalatt: I guess you can use `DiskManager.Read` when you'll refactor this code

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
alanmcgovern: It is typical for "count" to be equal to the buffer size, and that's how HttpStream itself is implemented. 

That said, I'm not sure what the issue is which led to your comment. Is there a mistake in the logic here?

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
alanmcgovern: Hah, that's a very good point! I should definitely fix that :) I suspect that would simplify things further too :)

This is something I did while prototyping in a standalone application and I never even thought to move to DiskManager! Oops :)

Great catch!

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
ilyalatt: Sure. You can read a data of the next file because you do not check `count`. Also you can set `Position >= Length` and read the other file data. Ok, it is not true, because you call `Stream.ReadAsync` and it should handle all this cases. But you download the data for this stream using `startPiece` and `endPiece`. So there are cases when it waits for redundant data. Maybe forever :)

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            // The torrent is treated as one big block of data, so this is the offset at which the current file's data starts at.
+            var torrentFileStartOffset = (long)File.StartPieceIndex * (long)Manager.Torrent.PieceLength + File.StartPieceOffset;
+
+            // Take our current position into account when calculating the start/end pieces of the data we're reading.
+            var startPiece = (int) (torrentFileStartOffset + Position) / Manager.Torrent.PieceLength;
+            var endPiece = (int) (torrentFileStartOffset + Position + count) / Manager.Torrent.PieceLength;
+            while (Manager.State != TorrentState.Stopped && Manager.State != TorrentState.Error) {
+                bool allAvailable = true;
+                for (int i = startPiece; i <= endPiece && allAvailable; i++)
```
ilyalatt: I guess there are other places with similar code. Maybe it should be moved to something like `bool BitField.AllInRange(int startPiece, int count)`? Name like `AnyInRange` for OR can be used too.

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
alanmcgovern: Ah yes, I see what you mean!

If too much data were requested then it would either be fine or an exception would be thrown if the endPiece doesn't exist. 

With the current implementation it's not possible to read data from a second file because only one FileStream is open and you won't be able to over-read. 

If this is swapped to DiskManager then overreading could be an issue so clamping the offset/count to the appropriate level would be the right thing to do. That's a great point

uponatime2019:
Sorry for late reply. This is my test
Theo video already downloaded to 100%

**When stream with HttpStream**
My code:
```
        private static async Task<IUriStream> CreateHttpStreamAsync(TorrentManager manager, TorrentFile file)
        {
            var provider = new StreamProvider(manager);
            return await provider.CreateHttpStreamAsync(file);
        }
```

Output: Seem the code loop like that and over when it finish read all the files, so UWP Media Player show progress  waiting infinity
```
Received HTTP request
bytes=0-:Client requested: {0}
Requested 0 -> 683208129. Length: 683208130
Read to 65536. HighPriorityPiece: 9 -> 9
Read to 131072. HighPriorityPiece: 9 -> 9
Read to 196608. HighPriorityPiece: 9 -> 10
Read to 262144. HighPriorityPiece: 10 -> 10
Read to 327680. HighPriorityPiece: 10 -> 10
Read to 393216. HighPriorityPiece: 10 -> 10
Read to 458752. HighPriorityPiece: 10 -> 10
Read to 524288. HighPriorityPiece: 10 -> 10
Read to 589824. HighPriorityPiece: 10 -> 10
Read to 655360. HighPriorityPiece: 10 -> 10
Read to 720896. HighPriorityPiece: 10 -> 11
Read to 786432. HighPriorityPiece: 11 -> 11
Read to 851968. HighPriorityPiece: 11 -> 11
Read to 917504. HighPriorityPiece: 11 -> 11
Read to 983040. HighPriorityPiece: 11 -> 11
Exception thrown: 'MonoTorrent.ConnectionClosedException' in MonoTorrent.dll
Exception thrown: 'MonoTorrent.ConnectionClosedException' in System.Private.CoreLib.dll
Read to 1048576. HighPriorityPiece: 11 -> 11
Exception thrown: 'MonoTorrent.ConnectionClosedException' in System.Private.CoreLib.dll
Read to 1114112. HighPriorityPiece: 11 -> 11
Read to 1179648. HighPriorityPiece: 11 -> 11
Read to 1245184. HighPriorityPiece: 11 -> 12
Read to 1310720. HighPriorityPiece: 12 -> 12
Read to 1376256. HighPriorityPiece: 12 -> 12
Read to 1441792. HighPriorityPiece: 12 -> 12
Read to 1507328. HighPriorityPiece: 12 -> 12
Read to 1572864. HighPriorityPiece: 12 -> 12
Read to 1638400. HighPriorityPiece: 12 -> 12
Read to 1703936. HighPriorityPiece: 12 -> 12
Read to 1769472. HighPriorityPiece: 12 -> 13
Read to 1835008. HighPriorityPiece: 13 -> 13
Read to 1900544. HighPriorityPiece: 13 -> 13
Read to 1966080. HighPriorityPiece: 13 -> 13
Read to 2031616. HighPriorityPiece: 13 -> 13
Read to 2097152. HighPriorityPiece: 13 -> 13
Read to 2162688. HighPriorityPiece: 13 -> 13
Read to 2228224. HighPriorityPiece: 13 -> 13
Read to 2293760. HighPriorityPiece: 13 -> 14
Read to 2359296. HighPriorityPiece: 14 -> 14
Read to 2424832. HighPriorityPiece: 14 -> 14
```

uponatime2019:
When Stream with LocalStream. Same above file
My code: 
```
private static async Task<Stream> CreateLocalStreamAsync(TorrentManager manager, TorrentFile file)
        {
            var provider = new StreamProvider(manager);
            return await provider.CreateStreamAsync(file);
        }
```

Output:
```
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 16. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 1024. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 1024. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 2048. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 4108. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 9. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 16. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 12. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 16. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 65536. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 4096. HighPriorityPiece: 9 -> 9
Seek to 0. HighPriorityPiece: 9 -> 9
Read to 262144. HighPriorityPiece: 9 -> 10
Seek to 683147264. HighPriorityPiece: 10 -> 1312
Exception thrown: 'System.ArgumentOutOfRangeException' in MonoTorrent.dll
Specified argument was out of the range of valid values.
```

And exception throw at BitField.cs L289
```
bool Get (int index)
        {
            if (index < 0 || index >= length)
                throw new ArgumentOutOfRangeException ("index");

            return (this.array[index >> 5] & (1 << (31 - (index & 31)))) != 0;
        }
```

alanmcgovern:
Can you share a copy of the media file and, if possible, a minimal uwp app which streams from HTTP?

The media file and uwp app I created work as expected. The file you have mist be exposing a bug in the logic which my file doesn't hit. 

If you can't provide those I'll try a few more random files and see what happens

uponatime2019:
> Can you share a copy of the media file and, if possible, a minimal uwp app which streams from HTTP?
> 
> The media file and uwp app I created work as expected. The file you have mist be exposing a bug in the logic which my file doesn't hit.
> 
> If you can't provide those I'll try a few more random files and see what happens

The torrent file is here
[temp1580705043.zip](https://github.com/alanmcgovern/monotorrent/files/4151416/temp1580705043.zip)

I can't provide minimal UWP app now, but I may try

src/MonoTorrent/MonoTorrent.Streaming/LocalStream.cs
alanmcgovern: This is actually part of a problem! When the UWP media player plays a file, it over-reads on purpose but it also does not actually expect/require it's read to be fulfilled. For example if the file is 100kb in size UWP might request bytes 90 -> 110, but if you only give it bytes 90->100 it will work as expected.

So yes - you were right and I need to figure out a good mechanism for clamping the data :)

alanmcgovern:
Thanks, that file was sufficient for me to figure out the issue. The gist of it is that i need to figure out how to provide data to an application which deliberately overreads :) ( https://github.com/alanmcgovern/monotorrent/pull/224#discussion_r374485110 ).

I'll push something relatively soon...

alanmcgovern:
@uponatime2019 When you use this method to stream the media and let me know if it works? If you pass one of these `StreamingRandomAccessStream` objects to the media player it should work with an unmodified monotorrent from this branch.

I'll try a few ways of moving the logic to proect against over-reads into MonoTorrent itself. It should be easy enough to add it in there :)

alanmcgovern:
Alrighty, i just added over-read protection to `LocalStream` and `HttpStream` so both of those *should* be safe to pass to UWP now. Hopefully this works for you!

alanmcgovern:
( https://recordit.co/I2e8wTQdFK )

uponatime2019:
I just test with latest commit and it's working fine with LocalStream, but HttpStream still same problem
I'm using this code to create Media Source
`MediaPlayer.Source = MediaSource.CreateFromUri(playLink.ToUri());`

alanmcgovern:
UWP has many ways to initialize the media player. The simplest approach I found to make it work was to not give it a Http url directly, but to wrap it in the StreamingRandomAccessStream class I created. This makes the data appear like a normal FileStream (more or less) and it Just Works. 

CreateFromUri appears to attempt to buffer the entire file into memory before it plays. Perhaps there's a HTTP trick I'm missing, but it's hard to tell. 

You couls try AdaptiveStreaming ( https://docs.microsoft.com/en-us/windows/uwp/audio-video-camera/adaptive-streaming#initialize-an-adaptivemediasource-using-httpclient ) or HSL, or whatever other magic ways the client supports. Some of those extra ways may require changes to monotorrents HttpStream to expose additional functionality. 

Patches are welcome if you wish to go that route. Otherwise the StreamingRandomAccessStream  should do the trick for you! If that doesn't work let me know. It'd what I've been using

uponatime2019:
> @uponatime2019 Can you use this method to stream the media and let me know if it works? ( https://gist.github.com/alanmcgovern/eb7481ca384f3903284be3e8dc609fdd ) If you pass one of these `StreamingRandomAccessStream` objects to the media player it should work with an unmodified monotorrent from this branch.
> 
> I'll try a few ways of moving the logic to protect against over-reads into MonoTorrent itself. It should be easy enough to add it in there :) Once it's added on that side then it should be possible to pass the built in LocalStream object directly to UWP instead of forcing it to bounce via this (really crappy) stream impl.

Sorry I miss this sample code. It's OK now but I will stay with LocalStream method ;)

alanmcgovern:
It probably worked if your file was small. 

I don't think I've managed to make UWP actually stream from a http source. I wasn't too bothered as it works for LocalStream and VLC, and after that I can accept patches or suggestions to support additional platform specific things. 

Pretty much any other transport can be layered on top of LocalStream or HttpStream, so platform specific things should be relatively easy. 

Thanks for testing the over-read fix!

uponatime2019:
So can you merge this to latest NuGet release? Let me know if you need any test

alanmcgovern:
There's still some pending fixes which have been pointed out in this PR, and there's a bunch of tests which need to be added before I'd feel comfortable shipping this in a supported release.

However i'm happy to ship it in an alpha nuget package. I'll upload an alpha release to here and nuget.

NOTE: The API should still be considered unstable. I might change this around a bit more before it's finalised.

alanmcgovern:
> 
> 
> So can you merge this to latest NuGet release? Let me know if you need any test

I was just wondering if you have been using the preview release. If so, has the support been working well enough for you?

uponatime2019:
> > So can you merge this to latest NuGet release? Let me know if you need any test
> 
> I was just wondering if you have been using the preview release. If so, has the support been working well enough for you?

Yes, I used it for a while but no problem

alanmcgovern:
I expect to merge this in the not too distant future! I'm still debating about the API I want to use to expose this.

Currently you have to create a TorrentManager, then wrap it in a `StreamProvider`, then use the `StreamProvider` to get a `Stream` for a file in that `TorrentManager`.

The problem is then:
What should the StreamProvider/Stream do when:
1. The TorrentManager is not running when the Provider is created. Auto start it? What if the torrent was only 
2. happens if someone starts/stops the TorrentManager o

alanmcgovern:
I'll keep it simple for now and just go with this. It shouldn't be too hard to change (if needed).

alanmcgovern:
This has been merged but it is not public yet. I'll make an issue listing the remaining work. I feel that the current impl is good enough to merge in, and everything else can be an incremental tweak on top of this :)

alanmcgovern:
Quick update on this - i'm making a few decisions and making the implementation public today or tomorrow: https://github.com/alanmcgovern/monotorrent/issues/257

PR #225, opened 2020-01-27
enhance disk manager performance

ilyalatt:
* replaces O(N) file search with O(log N) search
* adds a primitive cache

Should be tested in real world

### DIFF
--- a/src/MonoTorrent.Tests/Client/DiskManagerTests.cs
+++ b/src/MonoTorrent.Tests/Client/DiskManagerTests.cs
@@ -13,10 +13,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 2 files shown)


alanmcgovern:
Do you have an example torrent with lots of files which could be used for prototyping? Even a description which would allow me to recreate a reasonable imitation of it using local files would be good!

ilyalatt:
Sure. I have found a good one, https://rutracker.org/forum/viewtopic.php?t=4749512. It has 5001 files.

alanmcgovern:
I have an alternative patch which (i think) provides the same benefit with less changes! Could you take a look at https://github.com/alanmcgovern/monotorrent/pull/230 and see if it's as good in your use case? It makes a fairly large difference to the profiler results when testing with the torrent you provided :)

ilyalatt:
Looks good, but what about caching? It should help a lot.

alanmcgovern:
I'm not exactly sure what your patch is caching. Am i right in thinking that the idea is to allow faster lookups by storing the result of the previous lookup and attempting to re-use it? If that's the case then it's unlikely to provide any benefit in real world usage as torrents typically 16kb chunks randomly. Even if you are performing a sequential download, you should still have multiple peers downloading concurrently which would, in essence, give you the same 'random' ordering, except the random ordering would be over a range of 10-20 pieces instead of the whole torrent.

As such I don't think further caches would add any measurable improvement? Is that a fair thing to say, or have I misinterpreted your intentions?

alanmcgovern:
I'll add the other patch as-is for now, and additional caching can be added in a future patch as needed :) If we can show an improvement in some aspect of the library using the 5000+ file torrent you shared, then I'd be happy to include it!

ilyalatt:
@alanmcgovern I guess it should help when you calculate hashes of downloaded files. I checked the cache on tests and there was a lot of cache hits. Of course it is needed to be enhanced to be useful for more then one active torrent, like `ConcurrentDictionary<ITorrentData, int>`.

alanmcgovern:
Oh that's a good point. I do think the cache would be most useful when hash checking. However I don't know if the change in performance would be enough to warrant specialised caching for this.

If you compare these two charts you can see the relative performance. The original code was quite bad when you read data from the last piece of the torrent 11,000 times in a row:



The revised version of the code using binary search, as you suggested, is much better:



To put this into relative terms, the total CPU time needed to read data from the *last* piece used to be 28 seconds and it is now approximately 400ms. Out of that 400ms, about 360ms is actual disk IO.

If we added a layer of caching to the new codepath, the best-case we can save is somewhere between 0 and 40 milliseconds, which is less than 8% of the total time. After your first optimisation, there's very little to be gained with a cache :) The simplicity we have with a basic binary search is better than adding caching and saving less than 6-7% of the worst-case-scenario (linear hashing) and ~% of the standard (random downloading) scenario :)

If you have data to show otherwise i'll definitely reconsider, but for now I think it's probably safe to close this and thank for you for discovering and fixing the worst of this :D

PR #228, opened 2020-01-27
apply resharper global refactorings

ilyalatt:


alanmcgovern:
Awesome. I was just wondering which refactors you applied to the project, or was it just the default ones? Is this something that can be enabled in a config so that freshly added code abides by the same codestyle?

I'm relatively new to resharper :)

src/MonoTorrent.Tests/Common/TorrentTest.cs
```diff
@@ -92,14 +92,12 @@ private BEncodedDictionary CreateInfoDict ()
         private BEncodedList CreateFiles ()
         {
             BEncodedList files = new BEncodedList ();
-            BEncodedDictionary file;
-            BEncodedList path;
 
-            path = new BEncodedList {
+            BEncodedList path = new BEncodedList {
```
alanmcgovern: @ilyalatt Most modern style guides would recommend constructs like this be declared as a  `var` . Can resharper make this 'obvious' things var instead of using the full type?

src/MonoTorrent/MonoTorrent.Client.Messages.Libtorrent/LTMetadata.cs
```diff
@@ -91,15 +82,15 @@ public LTMetadata (byte extensionId, eMessageType type, int piece, byte[] metada
             : this ()
         {
             ExtensionId = extensionId;
-            this.messageType = type;
-            this.metadata = metadata;
-            this.piece = piece;
+            this.MetadataMessageType = type;
```
alanmcgovern: the `this.` is unnecessary now. Can a refactoring remove those, and a formatter setting ensure they stay removed?

src/MonoTorrent.Tests/Common/TorrentTest.cs
ilyalatt: There are cases when a parameter name equals to field name so `this.` can not be removed. You can convert private field names from camelCase to _camelCase format in resharper.

src/MonoTorrent/MonoTorrent.Client.Messages.Libtorrent/LTMetadata.cs
ilyalatt: There are cases when a parameter name equals to field name so this. can not be removed. You can convert private field names from camelCase to _camelCase format in resharper.

ilyalatt:
> Awesome. I was just wondering which refactors you applied to the project, or was it just the default ones? Is this something that can be enabled in a config so that freshly added code abides by the same codestyle?
> 
> I'm relatively new to resharper :)

I applied only part of available refactorings that are automatically suggested. You can use [ReSharper CLI](https://www.jetbrains.com/resharper/features/command-line.html). Aside from that I recommend you to try [Rider](https://www.jetbrains.com/rider/). It is based on ReSharper backend. It is such a pleasure to work with this IDE in comparison with VS or VS Code with OmniSharp. It is a big pain to develop in C# in those editors after you get used to Rider :)

ilyalatt:
You click Code -> Inspect Code...

image

and then...

image
image
image

For many issues you can run global refactorings:

image
image
image
image
image
image

ilyalatt:
I recommend to try to remove .editorconfig temporarily (while you refactor the code) because it seems like Rider does not suggest some refactorings like `this.` omitting if it present.

ilyalatt:
In addition to free trial (30 days) you can obtain [a free license for open source projects](https://www.jetbrains.com/community/opensource/).

alanmcgovern:
I actually have a jetbrains opensource license already! I just haven't used rider yet. I'd be curious how it's been progressing as the last time i used it seriously was a year or two ago.

I've squashed up this commit, and performed a few other refactorings to clean up the 'this.' stuff and a few other minor things. I'll squash the whole thing together to get a smaller overall diff, and so there's only 1 commit to skip past when traversing history.

Thanks for the initial PR!

alanmcgovern:
Merged as part of https://github.com/alanmcgovern/monotorrent/pull/229

PR #230, opened 2020-01-29
Diskmanager binarysearch

alanmcgovern:


alanmcgovern:
Original perf:

old

New perf:
new

The testcase was loading a torrent with 5000 files and attempting to read data from the last part of the last file. This was the 'worst case' scenario for the old logic.

alanmcgovern:
This is a variation of https://github.com/alanmcgovern/monotorrent/pull/225 and was insprised by @ilyalatt . I believe it's more or less the same idea, except with slightly fewer changes to the code.

src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
```diff
@@ -47,6 +47,32 @@ public class DiskManager : IDisposable
...
+            long PieceLength { get; }
+
+            public TorrentFileFinder (long offset, int pieceLength)
+            {
+                Offset = offset;
+                PieceLength = pieceLength;
+            }
+            public int Compare (TorrentFile x, TorrentFile y)
+            {
+                var fileStart = (x.StartPieceIndex * PieceLength) + x.StartPieceOffset;
```
Therzok: Might be worth documenting that y is never used.

src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
```diff
@@ -450,23 +476,22 @@ void ProcessBufferedIO (bool force = false)
             }
         }
 
+        internal int FindFileIndex (TorrentFile[] files, long offset, int pieceLength)
+        {
+            return Array.BinarySearch (files, null, new TorrentFileFinder (offset, pieceLength));
```
Therzok: So we're looking for `y` in the comparer which is never used.

src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
```diff
@@ -47,6 +47,32 @@ public class DiskManager : IDisposable
...
+            long Offset { get; }
+
+            long PieceLength { get; }
+
+            public TorrentFileFinder (long offset, int pieceLength)
+            {
+                Offset = offset;
+                PieceLength = pieceLength;
+            }
+            public int Compare (TorrentFile x, TorrentFile y)
```
Therzok:
```csharp
public int Compare (TorrentFile x, TorrentFile _)
```

src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
alanmcgovern: Yeah, unfortunately .NET doesn't have a good built-in API for this sort of thing. It really wants you to use `BinarySearch` as a method for inserting objects. It doesn't really want you to use it for *finding* objects :(

I could clean it up as an extension method which essentially gives you:

`BinarySearch (T[] array, Func<T, int>)` and then you pass a Func which implements the required logic.

I do use the same trick here ( https://github.com/alanmcgovern/monotorrent/blob/master/src/MonoTorrent/MonoTorrent.Client.PiecePicking/StandardPicker.cs#L184 ) so perhaps it makes sense to make it an extension method now?

Therzok: Insertion can be done with bitwise negation though.

See [here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.binarysearch?view=netframework-4.8#System_Collections_Generic_List_1_BinarySearch__0_) at examples:

> The BinarySearch(T) method overload is then used to search for two strings that are not in the list, and the Insert method is used to insert them. The return value of the BinarySearch(T) method is negative in each case, because the strings are not in the list. Taking the bitwise complement (the ~ operator in C# and Visual C++, Xor -1 in Visual Basic) of this negative number produces the index of the first element in the list that is larger than the search string, and inserting at this location preserves the sort order. The second search string is larger than any element in the list, so the insertion position is at the end of the list.

Therzok: Nevermind, I read your comment in reverse.

Therzok: Yeah, an extension method sounds good to me!

PR #233, opened 2020-01-30
simplify binary search

ilyalatt:


src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
```diff
@@ -450,27 +450,19 @@ void ProcessBufferedIO (bool force = false)
...
+        internal static int FindFileIndex (TorrentFile[] files, long offset, int pieceLength)
         {
-            static int Locator (TorrentFile file, ValueTuple<long, int> state) {
-                // Force these two be longs right at the start so we don't overflow
-                // int32s when dealing with large torrents.
-                long offset = state.Item1;
-                long pieceLength = state.Item2;
-
-                var fileStart = (file.StartPieceIndex * pieceLength) + file.StartPieceOffset;
+            int Comparator (TorrentFile file, long fileOffset) {
```
alanmcgovern: By making this an instance method and capturing state there will be additional memory allocations whereas previously there were none.  

I don't recall how frequently this is invoked but they could be non-trivial amounts of allocations. The purpose of the overload which takes a state parameter is to avoid these allocations :)

ilyalatt: Ah, ok, it is fixed.

alanmcgovern:
Changed the target branch so i could pull this changeset in to get CI to run :)

alanmcgovern:
(reopened at https://github.com/alanmcgovern/monotorrent/pull/235/files so CI runs)

PR #235, opened 2020-02-01
Simplify binary search

alanmcgovern:


src/MonoTorrent/MonoTorrent.Client/Managers/DiskManager.cs
```diff
@@ -450,27 +450,20 @@ void ProcessBufferedIO (bool force = false)
...
-
-                var fileStart = (file.StartPieceIndex * pieceLength) + file.StartPieceOffset;
+            static int Comparator (TorrentFile file, ValueTuple<long, int> offsetAndPieceLength) {
+                var (offset, pieceLength) = offsetAndPieceLength;
+                var fileStart = (long) file.StartPieceIndex * pieceLength + file.StartPieceOffset;
                 var fileEnd = fileStart + file.Length;
+                var leftCheck = fileStart <= offset;
+                var rightCheck = offset < fileEnd;
+                return !leftCheck ? 1
+                  : rightCheck ? 0
```
alanmcgovern: The main change here is to write the three if statements as nested ternary statements. I don't think this makes it noticeably *easier* to reason about. In addition, the new variable `leftCheck` is only used when it's negated, which is a little confusing too. It would be better if `leftCheck` were declared as `fileStart > offset` as that's the only way it's actually used.

I do like the explicit naming of `offsetAndPieceLength` instead of `state`, and the use of the tuple unwrapping, so I'll split this commit up and take that aspect of it! I'll retain the independent `if` statements as I do think they make things a little easier to reason about. It's really clear that `0` is returned only when the offset is within the bounds of the file, whereas with ternary it's not as clear.

PR #237, opened 2020-02-01
[core] Optimise memory for TorrentFile

alanmcgovern:
Rather than storing one bitfield per TorrentFile just
to represent the start/end piece, we can use a simple
ValueTuple. The bitfield is used as a selector to
simplify some operations, but a ValueTuple representing
the (inclusive) start/end piece can perform the same
job.

The typical usecase is when some files are marked as
'DoNotDownload' and the selector is used to exclude
those pieces in the PiecePicker.

### DIFF
--- a/src/MonoTorrent/MonoTorrent/BitField.cs
+++ b/src/MonoTorrent/MonoTorrent/BitField.cs
@@ -333,6 +333,13 @@ namespace MonoTorrent
             return this;
         }
 
+        internal BitField SetTrue (ValueTuple<int, int> range)
+        {
+            for (int i = range.Item1; i <= range.Item2; i++)
+                Set (i, true);
+            return this;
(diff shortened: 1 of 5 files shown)


alanmcgovern:
Original:

original

New:

new

that's a lot of memory savings as large bitfields are no longer stored in the TorrentFile.

src/MonoTorrent/MonoTorrent/BitField.cs
```diff
@@ -333,6 +333,13 @@ public BitField Set (int index, bool value)
             return this;
         }
 
+        internal BitField SetTrue (ValueTuple<int, int> range)
```
Therzok: I wonder why you're using ValueTuple instead of `(int, int)`

PR #238, opened 2020-02-01
[core] Prefer named tuple syntax over ValueTuple

alanmcgovern:
It allows naming the various parameters, which makes
it much easier to consume tuples throughout the codebase.

Suggestion from Therzok.


alanmcgovern:
@Therzok Here you go!

Issue #223, opened 2020-01-25
Starting, Stopping, and Hashing operations are blocking MainLoop


vlasenkoalexey:
Whenever one of the torrent managers are in Starting, Stopping, and Hashing state, others can't do any work.

From my experience if one of the torrents is in blocking state, system won't respond to adding new torrents, and existing torrents won't be able to change state until blocking operation completes.

As far as I understand current implementation, async/await operations are queued to MainLoop instead of ThreadPool as usual. So any long-running op, can block whole program.  Therefore those operations should be executed off MainLoop.

alanmcgovern:
Hrm, that doesn't make a whole lot of sense to me based off what I remember the code doing. StoppingMode should be completely asynchronous and doesnt have operations which block the thread it's executing on. 

Do you have any further details about what exactly is blocking? Perhaps there's some unintended/unforseen synchronous blocking in some network code which is supposed to be asynchronous? I believe some networking APIs resolve DNS synchronously even if you're using async methods and can cause issues like you describe.

alanmcgovern:
@vlasenkoalexey If you have any further information do let me know, otherwise I'll go ahead and close this issue as the current implementation is non-blocking and does allow one torrent to be started while a second one is stopping.

vlasenkoalexey:
I'll try to get a repro using profiler, or just extend sample app by adding relevant log information.
I clearly observe that behavior in my app, but I had to do a lot of changes to monotorrent library in order to make it work for UWP. But given my current understanding of how things work I believe that issue exists in core library.

alanmcgovern:
Perfect, thank you!

I suspect the problem will most likely be related to an asynchronous networking API which is not actually asynchronous. I should do a double check of the docs around socket/webrequest and see where DNS queries may cause the behaviour you describe.

alanmcgovern:
I added some new API to ensure that you can load `.torrent` files without blocking the thread.

https://github.com/alanmcgovern/monotorrent/commit/7a9e01b9cd06eeaa51bb7604f20e689bf1ec417b

I have not found anything which would block the main loop for more than 300ms or so. However, i have made some changes for some minor sources of mainloop blocking. Do let me know how your investigations go.

alanmcgovern:
I'll go ahead and close this for now. If you are still having issues with the latest release do let me know. The latest stable release has some modifications to explicitly move async tasks to threadpool before starting the operation. This should ensure that the library is protected from any hidden synchronous calls. 

[...]

PR #239, opened 2020-02-02
[core] Move some Web operations to the threadpool

alanmcgovern:
The profiler shows these as being potentially slow operations,
so ensure we bounce them to the threadpool so the synchronous
aspect of WebRequest.Create (and WebRequest itself) has no
impact on the main thread.

Partially address https://github.com/alanmcgovern/monotorrent/issues/223


alanmcgovern:
webrequest

It's only a small bit of CPU time.

PR #245, opened 2020-02-10
Faster bencoding

alanmcgovern:


src/MonoTorrent/MonoTorrent.BEncoding/BEncodeDecoder.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            BEncodedString key = null;
+            BEncodedValue value = null;
+            var torrent = new BEncodedDictionary ();
+            if (reader.ReadByte () != 'd')
+                throw new BEncodingException ("Invalid data found. Aborting"); // Remove the leading 'd'
+
+            int read;
+            while ((read = reader.ReadByte ()) != -1 && read != 'e') {
+                key = (BEncodedString) Decode (reader);         // keys have to be BEncoded strings
```
alanmcgovern: This looks like a bug. We're discarding the previous read.

src/MonoTorrent/MonoTorrent.BEncoding/BEncodeDecoder.cs
```diff
@@ -0,0 +1,165 @@
...
+
+                read = reader.ReadByte ();
```
alanmcgovern: Similar issue here?

alanmcgovern: Tests FTW!

src/MonoTorrent/MonoTorrent.BEncoding/BEncodeDecoder.cs
alanmcgovern: Tests FTW!

src/MonoTorrent/MonoTorrent.BEncoding/BEncodeDecoder.cs
```diff
@@ -0,0 +1,164 @@
...
+            while ((read = reader.ReadByte ()) != -1 && read != ':') {
+                if (read < '0' || read > '9')
+                    throw new BEncodingException ($"Invalid BEncodedString. Length was '{length}' instead of a number");
+                length = length * 10 + (read - '0');
+            }
+
+            if (read != ':')
+                throw new BEncodingException ("Invalid data found. Aborting");
+
+            var bytes = new byte[length];
```
Therzok: Are these strings really long? Otherwise maybe ArrayPool<byte> could be used here.

alanmcgovern: The byte[] is retained in the BEncodedString object forever. A BEncodedString is a raw byte[] which can sometimes be converted into a valid UTF-8 string. If a UTF-8 string is needed it is generated on the fly and not retained

Issue #246, opened 2020-02-11
Throw FormatException in MagnetLink.FromUri only if not a Tr record


amellini:
There's some cases when the magnet uri contain a malformed Tr record, now the FromUri throw a FormatException and cannot start the torrent.

I think tr record should be ignored and the process of validation should continue.
That's right?

(I'll make a PR with a fix)

alanmcgovern:
Your proposed fix has been merged to master! Thanks!

PR #247, opened 2020-02-11
Changed FromUri method in MagnetLink.cs to throw FormatExpception if …

amellini:
According to my issue #246  in this fix the FromUri will complete even a malformed TR record (wrong number of '=')


alanmcgovern:
Can you include a test which demonstrates the kind of magnet link this change will affect? On the face of it it's hard for me to tell the benefit of this change as there's no test showing it :)

amellini:
I've included a test, I know this is a "edge" case, but the afford to handle this is really small so I think this kind of situation can be handled.

alanmcgovern:
Perfect! Thanks for that. I edited the test just to remove references to real infohashes and tracker urls. It's probably best to not have those embedded in git history forever :)

I merged the adjusted test to master, along with your change. After reviewing it I can see there could be a benefit in *never* throwing exceptions when encountering unknown parameters in the query string, so i went ahead and made that change. Basically I removed the 'throw' and just left your new logic in there.

Thanks!

PR #251, opened 2020-02-24
[core] Reduce allocations invoking events async

alanmcgovern:
Save a fair few allocations when emitting a lot of
async events. Should reduce time spent in the GC in
some artificial benchmarks. Not sure if it'll make
a real world improvement though.


alanmcgovern:
Overall the 'dynamic' allocations are reduced by 38% while hashing. This was in a torrent with nearly 10,000 files and over 50,000 pieces.

profiling

PR #252, opened 2020-02-24
[core] Update bitfields faster when hashing

alanmcgovern:
When we're hashing torrents with hundreds or thousands of files
we can update the bitfield more efficiently if we use binary
search to find a matching file.

In addition we can bail out as soon as we find a file which no
longer contains the piece we're interested in, which is a saving
in and of itself.

Partially addresses https://github.com/alanmcgovern/monotorrent/issues/223


alanmcgovern:
Original CPU:

original_cpu

Final CPU:

first_optimization

PR #253, opened 2020-02-24
[core] Ensure crypto is not on the engine thread

alanmcgovern:
Push all the socket crypto work to a background thread.
While it's not a lot of CPU time, it's still stuff which
should be executed on the threadpool so the main loop is
available for syncing, and also to handle things like
StopAsync/StartAsync etc.

Partially addresses https://github.com/alanmcgovern/monotorrent/issues/223


alanmcgovern:
This screenshot shows 850ms of CPU time spent in various crypto related functions. This was over 10 start/stops of the same torrent manager, which means about 85ms of CPU time was spent running crypto on the main client engine thread every time i ran the torrentmanager for 10 seconds.

crypto_on_main_thread

While this is not necessarily problematic, it was unexpected. Move this to the threadpool so that the main loop is free!

no_crypto_main_thread

Issue #250, opened 2020-02-20
Two trackers


carlosib:
Hello!

I have some .torrent file all using two tracker. Some times I have the file complete only 1 tracker and incomplete in the other one. When I have it and the active tracker is incomplete the current .torrent file never is downloaded and it doesn't change to the tracker complete.

I did some test uisng uTorrent app and always do fast the download.

alanmcgovern:
I'm not sure exactly what the problem is from your description.

Are you saying that there are two trackers - one tracker has a fast peer and the other tracker had a slow peer. uTorrent queries both trackers, but MonoTorrent only queries one tracker. As a result uTorrent always gets the fast and slow peer but MonoTorrent might only get the slow peer?

carlosib:
Hello Alan McGovern!

I'll try to explain using others words.

My  torrent file has 2 tracers. Sometime when monotorrent load this .torrent file I have a tracker online and the other offline, but, monotorrent set the active tracker the offline and the active tracker is offline forever.

Sometime, also, I have 2 tracker and both online. But, for exemple, the tracker active is the tracker number 2 and this tracker the file is 'incomplete'. The tracker number 1 is online and it has the file 'complete'. But, I always have as active tracker the number 2.

Sorry my bad English!

alanmcgovern:
Hrmm, can you share the torrent so I can understand how the trackers are included?

If one tracker is offline then the second tracker should always be queried. 

If they are both online then it may be 'correct' to only query one tracker, depending on how they are included in the torrent.

I think I have a clear understanding of the problem you have (your English is great!).
 The only complication left is I don't have the .torrent file so I can't verify how the trackers are grouped into tiers :)

carlosib:
Hello!

Please, see the file like_momentotrailer_sobcustodia_231018.mp4.zip 

[like_momentotrailer_sobcustodia_231018.mp4.zip](https://github.com/alanmcgovern/monotorrent/files/4236810/like_momentotrailer_sobcustodia_231018.mp4.zip)

Thank you!

vlasenkoalexey:
I think I have an idea what this is all about, as I was thinking about changing the way TrackerManager currently works.

If torrent has 2 trackers, one of them has 1 peer, another one has plenty.
When download is started, trackers are randomized so it is not clear which one is going to be used first.

Logic in announce tries to send message to all trackers in the list sequentially, and it stops when it reaches first tracker that responds with peers.
[...]

PR #254, opened 2020-02-28
[core] Announce to all Tracker Tiers concurrently

alanmcgovern:
This is the intent of the spec. The fact MonoTorrent did not do
this before is a bug.

Fixes https://github.com/alanmcgovern/monotorrent/issues/250

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
@@ -47,27 +47,28 @@ namespace MonoTorrent.Client.Tracker
         /// <summary>
         /// Returns the tracker which will be used, by default, for Announce or Scrape requests.
         /// </summary>
+        [Obsolete("This is now a per-Tier value and should be accessed using TrackerTier.ActiveTracker.")]
         public ITracker CurrentTracker => Tiers.SelectMany (t => t.Trackers).OrderBy (t => t.TimeSinceLastAnnounce).FirstOrDefault ();
+
+#pragma warning disable CS0618 // Type or member is obsolete
+        TrackerTier CurrentTier => Tiers.FirstOrDefault (t => t.Trackers.Contains (CurrentTracker));
+#pragma warning restore CS0618 // Type or member is obsolete
 
         /// <summary>
         /// True if the most recent Announce request was successful.
         /// </summary>
-        public bool LastAnnounceSucceeded { get; private set; }
-
-        /// <summary>
-        /// The timer tracking the time since the most recent Announce request was sent.
-        /// </summary>
-        ValueStopwatch LastAnnounce;
+        [Obsolete ("This is now a per-Tier value and should be accessed using TrackerTier.LastAnnounceSucceeded.")]
+        public bool LastAnnounceSucceeded => CurrentTier?.LastAnnounceSucceeded ?? false;
 
         /// <summary>
         /// The time, in UTC, when the most recent Announce request was sent
         /// </summary>
-        public DateTime LastUpdated { get; private set; }
+        public DateTime LastUpdated => CurrentTier?.LastUpdated ?? DateTime.MaxValue;
 
         /// <summary>
         /// The TorrentManager associated with this tracker
         /// </summary>
-        ITrackerRequestFactory RequestFactory { get; set; }
+        ITrackerRequestFactory RequestFactory { get; }
 
         /// <summary>
         /// The available trackers.
@@ -77,7 +78,7 @@ namespace MonoTorrent.Client.Tracker
         /// <summary>
         /// The amount of time since the most recent Announce request was issued.
         /// </summary>
-        public TimeSpan TimeSinceLastAnnounce => LastAnnounce.IsRunning ? LastAnnounce.Elapsed : TimeSpan.MaxValue;
+        public TimeSpan TimeSinceLastAnnounce => CurrentTier?.TimeSinceLastAnnounce ?? TimeSpan.Zero;
 
         #endregion
 
@@ -92,7 +93,6 @@ namespace MonoTorrent.Client.Tracker
         internal TrackerManager (ITrackerRequestFactory requestFactory, IEnumerable<RawTrackerTier> announces)
         {
             RequestFactory = requestFactory;
-            LastAnnounce = new ValueStopwatch ();
 
             // Check if this tracker supports scraping
             var trackerTiers = new List<TrackerTier> ();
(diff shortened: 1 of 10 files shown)


alanmcgovern:
@carlosib Can you check this out and see if it solves your issue as expected? I'm going to tweak it a bit more to more officially deprecate some fields which are unreliable now that concurrent tier announcing/scraping is supported by the library.

Most of the fields on TrackerManager which show metadata about the most recent announce need to be moved to the `TrackerTier` instead.

carlosib:
Hello Alan McGovern!

Thank you! But, sorry, but, how can I do downloading of your source code?

alanmcgovern:
Ah, if you're not familiar with cloning, compiling and testing fixes then it may be simpler to wait for me to do a release.

However if you want to try then follow the github instructions for cloning. You can view them by clicking on the `command line instructions` link in this page.

instructions

One the repository has been cloned you can compile MonoTorrent.sln in Visual Studio (or equivalent).

Then reference the binaries in your project instead of referencing the NuGet package. Alternatively you could add MonoTorrent.csproj to your solution file and make your other projects reference that instead of the nuget package.

alanmcgovern:
I'd just like to add that there are still a lot of changes I can make to this experience to make it even better. The tricky part will be making this work without breaking the API. I have some thoughts but it'll take a few days before I have the time to implement all the changes.

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
```diff
@@ -108,80 +108,114 @@ internal TrackerManager (ITrackerRequestFactory requestFactory, IEnumerable<RawT
...
         public async Task Scrape ()
         {
-            await Scrape (null);
+            var scrapes = new List<Task> ();
+            for (int i = 0; i < Tiers.Count; i++) {
+                var tier = Tiers[i];
+                for (int j = 0; j < tier.Trackers.Count; j++) {
+                    var tracker = tier.Trackers[j];
+                    if (tracker.CanScrape && tier.TimeSinceLastScrape > tracker.UpdateInterval)
+                        scrapes.Add (Scrape (tracker));
```
vlasenkoalexey: Why this logic is different than announces?
Shouldn't scrape also be per tier?

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
```diff
@@ -108,80 +108,114 @@ internal TrackerManager (ITrackerRequestFactory requestFactory, IEnumerable<RawT
...
-                    if (!trackerTier.SentStartedEvent)
-                        actualArgs = actualArgs.WithClientEvent (TorrentEvent.Started);
-
-                    List<Peer> peers = await tracker.AnnounceAsync (actualArgs);
-                    LastAnnounceSucceeded = true;
-                    AnnounceComplete?.InvokeAsync (this, new AnnounceResponseEventArgs (tracker, true, peers.AsReadOnly ()));
+        async Task AnnounceToTier (TorrentEvent clientEvent, TrackerTier tier)
+        {
+            for (int i = 0; i < tier.Trackers.Count; i++) {
+                int trackerIndex = (i + tier.ActiveTrackerIndex) % tier.Trackers.Count;
```
vlasenkoalexey: Could you please add a comment, I'm not sure I'm following logic here on how index is calculated

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
```diff
@@ -108,80 +108,114 @@ internal TrackerManager (ITrackerRequestFactory requestFactory, IEnumerable<RawT
...
+
+                List<Peer> peers = await tracker.AnnounceAsync (actualArgs);
+                trackerTier.LastAnnounceSucceeded = true;
+
+                trackerTier.ActiveTrackerIndex = trackerTier.Trackers.IndexOf (tracker);
+                trackerTier.SentStartedEvent |= actualArgs.ClientEvent == TorrentEvent.Started;
+                trackerTier.LastAnnounce = ValueStopwatch.StartNew ();
+                AnnounceComplete?.InvokeAsync (this, new AnnounceResponseEventArgs (tracker, true, peers.AsReadOnly ()));
+                return true;
+            } catch {
```
vlasenkoalexey: May I make a suggestion. Currently you have following TrackerStates:
    public enum TrackerState
    {
        Ok,
        Offline,
        InvalidResponse
    }

Could you also add Unknown and connecting Connecting.
Looks like here is the right place to set tracker state now.
All trackers should start with Unknown state, once you start connecting, you change it to Connencting.
If connection succeed, you change it to Ok, if fails - to Offline or InvalidResponse.

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
alanmcgovern: Yeah, it should be! This will be changed soon enough!

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
alanmcgovern: Suppose there are 10 trackers in the `TrackerTier` and `tracker[4]` was the one which responded most recently. This logic ensures we try each tracker sequentially `trackerIndex` will be calculated as:
```
(0 + 4) % 10   // 4
(1 + 4) % 10   // 5
(2 + 4) % 10   // 6
(3 + 4) % 10   // 7
(4 + 4) % 10   // 8
(5 + 4) % 10   // 9
(6 + 4) % 10   // 0
(7 + 4) % 10   // 1
(8 + 4) % 10   // 2
(9 + 4) % 10   // 3
```

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
alanmcgovern: Yup, makes sense. That should be easy to add!

alanmcgovern:
> Unfortunately, using the file
> [like_series_succession_280220.mp4.torrent.zip](https://github.com/alanmcgovern/monotorrent/files/4270689/like_series_succession_280220.mp4.torrent.zip) monotorrent aren't to able to do download this file :(
> 
> But, If I open 'uTorrent' app the download is 'snapshot' and very fast.

I'll have to look into this some more, but according to wireshark the remote client is closing the connection as soon as the `HaveNone` message, and `AllowedFast` messages are received.

weird_behaviour

This doesn't make sense as everything looks correct up until that point. Odd :) If you can - please open a separate github issue for this as it's unrelated to trackers!

alanmcgovern:
Update: even utorrent cannot connect to some of those peers. I'd recommend updating them from MonoTorrent 1.0 to the latest release, which is currently 1.0.15, if you are in control of those clients.

src/MonoTorrent/MonoTorrent.Client/Managers/TrackerManager.cs
alanmcgovern: Tracking this here: https://github.com/alanmcgovern/monotorrent/issues/255

PR #265, opened 2020-03-07
Fixing ProtocolUp counter

vlasenkoalexey:
Currently ProtocolUp counter has negative value, and DataUp counter is also probably wrong


alanmcgovern:
I added tests covering this case here: https://github.com/alanmcgovern/monotorrent/pull/266

alanmcgovern:
Thanks for the patch! This was a good catch. It's kinda funny that the comment was correct but the code was wrong :p

PR #267, opened 2020-03-08
Propagate exceptions stopasync

alanmcgovern:


src/MonoTorrent/MonoTorrent.Client/Modes/StoppingMode.cs
```diff
@@ -77,6 +78,8 @@ public async Task WaitForStoppingToComplete (TimeSpan timeout)
                 var overallTasks = Task.WhenAll (stoppingTasks);
                 if (await Task.WhenAny (overallTasks, delayTask) == delayTask)
                     Logger.Log (null, "Timed out waiting for the announce request to complete");
+                else
+                    await overallTasks;
```
alanmcgovern: Add a test and let the exception propagate?

Issue #257, opened 2020-03-03
Enable StreamProvider by default for streaming media


alanmcgovern:
Disorganised brain dump:

I expect to merge this in the not too distant future! I'm still debating about the API I want to use to expose this. A lot of these ideas could wait til V2 :)

Currently you have to create a TorrentManager, then wrap it in a StreamProvider, then use the StreamProvider to get a Stream for a file in that TorrentManager. This is a bit confusing as it's unclear what the lifespan/ownership of the TorrentManager is. How does starting/stopping affect everything?

Other thoughts include:

    Should the StreamProvider 'own' the TorrentManager and start/stop it as appropriate? If so what is the right logic?
    What if the TorrentManager was created with a MagnetLink and doesn't have metadata?
    What happens if the user tries to create multiple streams for one TorrentFile.
    What happens if the user wants to stream multiple files simultaenously. Currently only 1 file will stream correctly.
    Suppose there was a completely different implementation of ClientEngine/TorrentManager which dealt exclusively with streaming, what would that look like? Would it be better?

If anyone has any thoughts on this do let me know! Currently i think the most likely outcome is that I'll clean up some of the edgecases with the current approach and then release it more or less as-is.

OneFingerCodingWarrior:
Create multiple streams for one TorrentFile or Stream multiple files simultaneously should not be allowed. The reason is file is still being downloading, and the peers which uploading the data are out of our control. If we insist to implementing these capabilities then we will have to implement a full functional media serving server.

OneFingerCodingWarrior:
TorrentManager should independent of StreamProvider,  TorrentManager  should do the general purpose as to downloading the media file. Because most of the time, the experience of  StreamProvider will not be as good as local file playback. if the user do not like the media file, he will cancel the downloading immediately

alanmcgovern:
@OneFingerCodingWarrior I agree with both of your suggestions!

I've added APIs to StreamProvider to attempt to hide the fact a TorrentManager is needed/used. The underlying TorrentManager is still exposed so you can state rate limits, or access information related to the download. However you do not, and should not, call 'StartAsync' or 'StopAsync' on the TorrentManager, nor will you have to Register/Unregister it with the ClientEngine.

The API is now something like:
```
var provider = new StreamProvider (clientEngine, saveDirectory, magnetLink);
// or
provider = new StreamProvider (clientEngine, saveDirectory, torrent);

await provider.StartAsync ();

[...]

PR #271, opened 2020-03-10
Enable media streaming

alanmcgovern:


alanmcgovern:
Should address  https://github.com/alanmcgovern/monotorrent/issues/257 well enough for now

alanmcgovern:
@OneFingerCodingWarrior I've (hopefully) simplified things and am in a place where I'm comfortable shipping this in the next release!

I would love it if you could give this new API a shot and let me know if it works well for you. If you have any suggestions to make it easier to work with let me know! If it fails miserably then I really want to know what went wrong :p

alanmcgovern:
@ilyalatt @uponatime2019 Can you take a read my previous comment on this PR. If would be great if you could test things and let me know how you get on. I'd love to release this as long as there are no show-stopper issues! I won't make breaking changes to this API until i release a 'MonoTorrent 2.0.0' :)

OneFingerCodingWarrior:
Well , I will try the new API soon or later.
In fact I also implemented a simple http media streaming function using wcf, but because my code style (well ,you have met it ^_^) and dependency of WCF conflict with MonoTorrent, and unfamiliar with git, so I didn't push the changes. But  I am pretty happy to switch to your new API, and thanks for all the work you have done,

alanmcgovern:
> In fact I also implemented a simple http media streaming function using wcf

My hope is that anyone who wants to expose the data in a custom service/whatever can simply call StreamProvider.CreateStream and then read any data they need from that! They will be able to seek that stream and read data from it and MonoTorrent will magically ensure that data is prioritised/downloaded!

alanmcgovern:
I went ahead and merged this! I'll test it over the next few days before issuing a release with this! There is a small chance I'll make some tweaks between now and then, but i'd say it's unlikely :)

PR #278, opened 2020-03-18
Port forwarding

alanmcgovern:
Integrate port forwarding with MonoTorrent.

Through the magic of NuGet this is no longer a pain to support!

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client.PortForwarding/IPortForwarder.cs
+++ b/src/MonoTorrent/MonoTorrent.Client.PortForwarding/IPortForwarder.cs
@@ -1,0 +1,89 @@
+﻿//
+// IPortForwarder.cs
+//
+// Authors:
+//   Alan McGovern <alan.mcgovern@gmail.com>
+//
+// Copyright (C) 2020 Alan McGovern
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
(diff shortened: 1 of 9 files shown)


alanmcgovern:
@carlosib This build has built-in support for NAT-PMP and UPNP port forwarding. This might help with your connectivity issues if the machines are behind a router which has upnp/nat-pmp enabled.  If you do try this, do let me know if it helps.

[MonoTorrent.1.0.19-alpha.zip](https://github.com/alanmcgovern/monotorrent/files/4346323/MonoTorrent.1.0.19-alpha.zip)

Also, a good way to check if your bittorrent client is publicly reachable is to use this website: https://www.canyouseeme.org , or any similar port checking website. If you go to that website on the computer which is running your torrent client, and then put in the port, it will tell you if the port is accessible or not. This is the same as the 'telnet' test I was running before, but this is easier to use as it doesn't require multiple computers.

alanmcgovern:
@carlosib I have an update to this. I'm nearly ready to put something on nuget, but I want to think through one or two more problems first.

The difference is that port forwarding is not enabled by default anymore. You must call `ClientEngine.PortForwardingEnabled = true`. You can call it before, or after, starting a download. You can also turn it off any time you need. I attached a compiled verison of it here which you can use for testing.

[MonoTorrent.1.0.19-beta.zip](https://github.com/alanmcgovern/monotorrent/files/4390806/MonoTorrent.1.0.19-beta.zip)

Issue #291, opened 2020-04-19
Separate torrent download exception


justscribe:
Hi,

For some reason I'm getting error while try to start a second torrent, when the first is already downloading. Version is 1.0.22-master+733ad248.
Stacktrace:
19.04.2020 22:44:14> Value was invalid. | System.ArgumentException: Value was invalid.
   at System.Security.Cryptography.HashAlgorithm.TransformBlock(Byte[] inputBuffer, Int32 inputOffset, Int32 inputCount, Byte[] outputBuffer, Int32 outputOffset)
   at MonoTorrent.Client.DiskManager.<GetHashAsync>d__55.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.ReusableTaskAwaiter`1.GetResult()
   at MonoTorrent.Client.Modes.HashingMode.<WaitForHashingToComplete>d__15.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MonoTorrent.Client.TorrentManager.<HashCheckAsync>d__171.MoveNext() | 
19.04.2020 22:44:14> Value was invalid. |

OneFingerCodingWarrior:
to narrow down the issue,  try replace the ReusableTask with Task, and verify the result

alanmcgovern:
@justscribe can you share the two torrents?

justscribe:
1. https://mega.nz/file/woBBEQqL#wOYIGVlUVxZm918Bs_t_dFHB7YTYiLzXgVd33xW-aEc
2. https://mega.nz/file/AlBjTARa#m_KmH6RZmk8NqVgsiQ32-on0O9hz6QO0bI4K_o3kzEo

alanmcgovern:
Can you let me know if this fixes the issue? https://github.com/alanmcgovern/monotorrent/pull/292 ?

If not, can you reproduce with a debug build from that branch? If the issue reproduces can you share the details from the exception so we can tell the offsets, array length, etc from the TransformBlock call which throws an exception? I can't repro your issue locally, but i did find a related problem which *might* be what you're seeing.

justscribe:
Hi,

Thanks, fix is working.
[...]

PR #292, opened 2020-04-20
[core] Ensure data is not mixed between torrents

alanmcgovern:
Ensure the incremental hasher takes the torrent manager
as well as the piece index into account.

Ensure we don't have unexpected interaction between
multiple concurrent downloads.


alanmcgovern:
This was confirmed to fix https://github.com/alanmcgovern/monotorrent/issues/291#issuecomment-616755979

PR #301, opened 2020-05-22
adding tracker urls when torrent was loaded from magnet link

vlasenkoalexey:


alanmcgovern:
This looks great to me! I'll add a test to verify this works :)

alanmcgovern:
Merged manually after I added a test covering this change!

Your commit: https://github.com/alanmcgovern/monotorrent/commit/27b8ea5fd1c85ec5a03ce248348b83f94a5c3fac

My tests: https://github.com/alanmcgovern/monotorrent/commit/7400a9bbaa5fdde13dd5c28dc8e3b81a8e3e6e17

alanmcgovern:
Note: I did change the final implementation so the AnnounceUrls aren't exposed in two places. They're already accessible via the `TrackerManager` :)

PR #303, opened 2020-05-25
[core] Properly handle bitfield messages in metadata mode

alanmcgovern:
We should ignore them when they're received. This is
implemented by special casing the situation where a
BitfieldMessage is received while we do not have
metadata - in this scenario we return the special value
'BitfieldMessage.UnknownValue' which we can check for
later.

If we attempt to *send* this to a peer it will error
out when you invoke 'Encode'. The MetadataMode overload
for 'HandleBitfield' asserts that all bitfields were
decoded as this sentinal value.

Before this patch we could only retrieve metadata from
peers which have no data, or from seeders. With this
patch we can download metadata from leechers too.

### DIFF
--- a/src/MonoTorrent.Tests/MonoTorrent.Client.Modes/MetadataModeTests.cs
+++ b/src/MonoTorrent.Tests/MonoTorrent.Client.Modes/MetadataModeTests.cs
@@ -34,6 +34,7 @@ using System.Threading.Tasks;
 
 using MonoTorrent.Client.Encryption;
 using MonoTorrent.Client.Messages;
+using MonoTorrent.Client.Messages.FastPeer;
 using MonoTorrent.Client.Messages.Libtorrent;
 using MonoTorrent.Client.Messages.Standard;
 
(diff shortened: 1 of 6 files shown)


src/MonoTorrent.Tests/MonoTorrent.Client.Modes/MetadataModeTests.cs
```diff
@@ -112,12 +113,36 @@ public async Task RequestMetadata ()
             Assert.AreEqual (rig.Torrent.InfoHash, new InfoHash (new SHA1Managed ().ComputeHash (stream)), "#1");
         }
 
+        [Test]
+        public async Task AfterHandshake_SendBitfieldMessage()
+        {
+            var torrent = Path.Combine (AppDomain.CurrentDomain.BaseDirectory, "file.torrent");
+            await Setup (true, torrent);
+            await SendMetadataCore (torrent, new BitfieldMessage (rig.Torrent.Pieces.Count));
```
alanmcgovern, line 121: @vlasenkoalexey This should solve your problem. This means that monotorrent can retrieve metadata from leechers in all circumstances. Without this patch it could only download metadata from seeders or leechers which have not downloaded any data. The tests, unfortunately, didn't exercise that case.

PR #307, opened 2020-06-02
accepting block even if hash check failed for other peer

vlasenkoalexey:
When I debugged picker logic I got to this line.
I'm not sure that I understand why it is necessary to skip block when hash check fails for some other random peer. I think this is not necessary.
Of course I might be missing something, please take a look.

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client.PiecePicking/StandardPicker.cs
+++ b/src/MonoTorrent/MonoTorrent.Client.PiecePicking/StandardPicker.cs
@@ -239,8 +239,7 @@ namespace MonoTorrent.Client.PiecePicking
                 Piece p = requests[pIndex];
                 // If the peer who this piece is assigned to is dodgy or if the blocks are all request or
                 // the peer doesn't have this piece, we don't want to help download the piece.
-                if (p.AllBlocksRequested || p.AllBlocksReceived || !peer.BitField[p.Index] ||
-                    (p.Blocks[0].RequestedOff != null && p.Blocks[0].RequestedOff.RepeatedHashFails != 0))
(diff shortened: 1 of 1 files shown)


alanmcgovern:
tl;dr is that sometimes clients can accidentally have corrupt data. This check is to ensure that clients who send corrupt data can be detected and banned. 

If you're connected to 3 peers and they are all helping download the same piece, we mark all peers as having sent a corrupt piece if that piece fails it's hash check. 

The next time we request data from each of those peers we require that they each send all the data for a complete piece by themselves. This means we can easily tell if one of them is consistently sending corrupt data. If a piece fails a hashcheck at this point then we know exactly which peer sent the data and can mark them as having sent corrupt data.

If the piece passes the hashcheck then we mark those peers as having RepeatedHashFails = 0 so they are allowed  help download the same piece as another peer again.

In other words - we mark a peer as unhealthy if they are involved in a piece which fails the hashcheck, and we mark them healthy if they send a complete piece, all by themselves, which passes the hash check. 

Finally, sometimes peers randomly send corrupt data. You protect against this we always close the connection for any peer which is involved with 5 different hash fails, whether they sent all the data, or part of the data, for each of those pieces. 

RepeatedHashFails measures the number of times in a row a peer sends corrupt data. TotalHashFails is the overall total.

alanmcgovern:
I'll close this as this is an intentional part of the logic in the piece picker. The aim is to ensure that peers which are sending corrupt data end up corrupting as *few* pieces as possible. It's better to allow them to send all the blocks for 1 piece and corrupt that, rather than allow them to send 1 block from 16 pieces and corrupt all of them :)

PR #317, opened 2020-07-06
Significant refactorings to enable improved performance

alanmcgovern:
Significant refactoring to make the library more async internally. This allows a significant boost in performance for seeding and leeching. This may impact actual real world download performance in some scenarios, but most likely gain will be a slight reduction in both CPU and memory usage while using the library.

Before:
old

After:
new

This test is performed using the 'StressTest' class in the repository. It instantiates one instance of MonoTorrent and uses it as a seeder. It then starts 16 more engines and puts them in an infinite loop where they download the data, discard it, and download again. The stress test does not write any data to disk, as the load from the random reads/writes from over 250 concurrent connections (16 leechers connecting to each other) would likely the limits of any SSD.

This is a fairly significant API break.


vlasenkoalexey:
Wow, this is amazing, tons of great stuff!!!
FYI: When I tried to make IO async, I couldn't make it work reliably without locking around stream and using sync version on stream IO: https://github.com/vlasenkoalexey/monotorrent/commit/7c48291692093cad64039d9eedafe09efdf37605#diff-a4b4af14c838d67859819ef101c1368cR96-R106
Without those tweaks download stuck at 95-99% for large files due to some hash misses.
Hope your stress test covers that.
In any case, thanks for doing all those changes. I'm super eager to try them out.

alanmcgovern:
I did find a few small issues when setting up the stress test. As it stands the stress test really forces the engine to repeatedly stop/start/reconnect so things should be quite robust now.  It made it so easy to profile :)

There are still a few lazy Jack's unfortunately. I left a SemaphoreSlim in the TorrentFile class, but it should really be elsewhere. I'll do a few more cleanups to tidy up the refactor, but then this leaves the library in a place where the work to support UWP streams will be easier to integrate :)

alanmcgovern:
Time to YOLO this into master! After the profiling and terabytes of data I've transferred between instances using the new StressTest.cs, i'm fairly confident things have not regressed too much :)

I won't release a '2.0' release for another while as there are more things I want to get into the release. However it's about time to get it into master so i'm not dragging around a massive diff forever :)

PR #319, opened 2020-07-14
[core] Hide 'TorrentFile' behind an interface and make it immutable

alanmcgovern:
The main goal of this is to make ITorrentFile and Torrent immutable.
It also clearly separates the mutable state, like the semaphore
used to access the underlying datastore in a threadsafe way, from
the data we parse from the actual .torrent metadata.


alanmcgovern:
NOTE: There's a bug in this refactor which causes the new stress test to not download reliably.

Issue #322, opened 2020-08-11
Wrong value of LocalPeerDiscovery.BroadcastEndPoint property 


vstrizhakov:
The BroadcastEndPoint in LocalPeerDiscovery seems to be wrong because it's current ip-address is IPAddress.Broadcast, but if I'm not wrong, it must be the 239.192.152.143, as specified in BEP 14. At least, if I use the IPAddress.Broadcast one, my udp-receiver (on another machine) doesn't receive the multicast message. And with 239.192.152.143 ip-address everything work like it need to be.

alanmcgovern:
Your patch has been merged so i'll go ahead and close this! Thanks for submitting the PR!

PR #323, opened 2020-08-12
issue#322 Using MulticastIpAddress instead of IPAddress.Broadcast for initializing UdpClient in LPD

vstrizhakov:
The description of a problem is [here](https://github.com/alanmcgovern/monotorrent/issues/322)


alanmcgovern:
Huh, strange. This seemed to work in the past with uTorrent. Good catch!

alanmcgovern:
Thanks for the patch. I'll backport this to 1.0 as well

PR #325, opened 2020-08-14
Fixed the issue when Torrent.Source stays null when torrent was loaded using Torrent.Load(string path) method

vstrizhakov:
These lines were replaced in 6a32c0c0421a05054361711ae41108821234988a commit, but I think this was done by mistake, because now Torrent.Source is always null even if loading torrent from file.

### DIFF
--- a/src/MonoTorrent/MonoTorrent/Torrent.cs
+++ b/src/MonoTorrent/MonoTorrent/Torrent.cs
@@ -560,7 +560,9 @@ namespace MonoTorrent
             Check.Path (path);
 
             try {
-                return LoadCore ((BEncodedDictionary) BEncodedValue.Decode (stream));
+                var torrent = LoadCore ((BEncodedDictionary) BEncodedValue.Decode (stream));
+                torrent.Source = path;
+                return torrent;
             } catch (BEncodingException ex) {
                 throw new TorrentException ("Invalid torrent file specified", ex);
             }


alanmcgovern:
Ah, i think there's a misunderstanding. There used to be a property called `TorrentPath` and a property called `Source`. Now there is only a property called `Source`. The `Source` property exposes the value of the 'Source' key from the torrent metadata, if it is set. Typically this value is related to the person/company/group who created the torrent.

The old code used `Torrent.TorrentPath` to store the filepath where the `Torrent` instance was originally loaded from. As part of the refactor you indicated, this property was removed completely. The intention is that the `Torrent` object will be completely immutable. A second goal is that the `Torrent` instance you end up with will be identical as long as you load it using the same underlying metadata.

For example, a `Torrent` object loaded using: `Torrent.Load ("path/to/file.torrent")` is identical to that loaded using `Torrent.Load(File.ReadAllBytes("path/to/file.torrent"))`.

If the filepath is part of the `Torrent` instance, then this would not be possible. The master branch is still unstable as I refactor things and make more API breaking changes, so it's possible to revisit this decision. At the moment I think the 'be immutable and idempotent' goal has been positive and has allowed me to properly address some thread safety issues :)

vstrizhakov:
Ahhh, yes, it’s a completely misunderstanding, my bad. Thank you for the explanation :)

PR #327, opened 2020-09-17
Mono.Nat package update fix.

justscribe:
App is crashing after Mono.Nat was updated. Previous implementation had a DeviceLost event, that is was removed.
Exception: System.MissingMethodException: Method not found: 'Void Mono.Nat.NatUtility.add_DeviceLost(System.EventHandler`1<Mono.Nat.DeviceEventArgs>)'.


alanmcgovern (approved):

alanmcgovern:
Lgtm!

Thanks for pushing this!

justscribe:
Nice=)

PR #335, opened 2020-11-11
Fix moving unfinished, stopped torrent.

doMConSwiss:
Moving files of even a stopped torrent will crash if a file does not yet exist. This checks for existence first and does not move the not existing file.


alanmcgovern (approved):
Awesome. This looks good to me!

Issue #320, opened 2020-08-06
StreamProvider Torrent.Files is always null.


Kajirinn:
Hello there, this is probably a mistake on my part but I can't figure it out for the life of me. I've been trying to find documentation about the StreamProvider, or the library in general; but have been having a rough time.

I'm trying to develop a simple video streaming application in .NET Framework so I don't have to fully download torrents before watching them. I've worked a bunch with vlcdotnet before so that isn't necessarily a problem, it's just I'm currently stuck on the `CreateHttpStreamAsync()` as the torrent files always return null.

I've been trying to figure it out through bits and pieces of codes from closed issues, such as #315 and #294, but it's sadly not working out for me.

```
MagnetLink MagnetLink = MagnetLink.Parse(magnet);
ClientEngine Engine = new ClientEngine();

StreamProvider StreamProvider = new StreamProvider(Engine, @"D:\Temp\test.torrent", MagnetLink, @"D:\Temp\meta.torrent");
await StreamProvider.StartAsync();
await StreamProvider.WaitForMetadataAsync();

Stream stream = await StreamProvider.CreateStreamAsync(StreamProvider.Manager.Torrent.Files[0]);
```

I don't think I'm understanding it correctly, why would the streamprovider need the torrent file itself, when it only recently 'started' with the magnet link I provided? Why is it still null after it supposedly waited for the metadata? The magnet link is valid as a client such as BitTorrent is loading it in correctly.

Besides this issue, are there any starting guides or simple sample projects? I peeked a little in the sample projects I've found, but either they're overwhelmingly big for the simple task that I want to achieve, or out of date.

Forgive my ignorance, and thanks in advance.

Kajirinn:
Upon messing around with it a tad bit more, I realized that the dir's don't want extensions. The `WaitForMetadataAsync()` now populates the `StreamProvider.Files`, instead of the `StreamProvider.Manager.Torrent.Files`. This generates a metadata torrent if I understand it correctly, but still throws an error on the `CreateStreamAsync()`.

```
Exception thrown: 'System.NullReferenceException' in mscorlib.dll
System.NullReferenceException: Object reference not set to an instance of an object.
   at MonoTorrent.Streaming.StreamProvider.<CreateStreamAsync>d__37.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #337, opened 2020-11-17
[core] Fix a couple of additional metadatamode races

alanmcgovern:
Firstly, don't reset the stream every time we connect
to a peer who supports metadata mode. That's... silly.

Secondly, once we get the validated metadata we should
emit the 'MetadataReceived' event and bail.

Thirdly, once we have validated metadata we can safely
ignore any freshly received LTMetadata messages as we
know the data is no longer needed.

Each of these changes deserves a test :p


alanmcgovern:
Fixes the remaining issues so that i can download ubuntu using metadata mode.

Should also address the issues from https://github.com/alanmcgovern/monotorrent/issues/320

PR #338, opened 2020-11-20
fix local peer discovery on systems with multiple network interfaces

mr-const:
Added code, which joins multicast group on each network interface with IP address, this resolves case, when you seed from computer with multiple network cards and leeches connect to you after long timeout (at least 5min)


alanmcgovern:
Great! This looks good to me. Thanks for the patch !

Issue #332, opened 2020-09-28
[ .torrent ] File parsing Error in Monotorrent, when added url-list (http webseed url) option in [ .torrent ] File, created from Tixati, Transmission Client


wizard872:
[ .torrent ] File parsing Error in Monotorrent, when added url-list (http webseed url) option in [ .torrent ] File, created from Tixati, Transmission Client

Tixati, Transmission

![image](https://user-images.githubusercontent.com/36080488/94421344-5e289a80-01c0-11eb-8eff-f61c0c71f02f.png)

but created from uTorrent, is work and not Parsing error.

uTorrent

![image](https://user-images.githubusercontent.com/36080488/94420831-ad220000-01bf-11eb-8d29-04422ce4b83d.png)

and i see .torrent file with HxD Hexeditor,

[HxD Hexeditor](https://mh-nexus.de/en/hxd/)

.torrent parsing ok, utorrent added url-list options, end of .torrent file like this,

![image](https://user-images.githubusercontent.com/36080488/94421584-bb245080-01c0-11eb-9ae3-e62ea54dcad7.png)

but .torrent parsing error  Tixati, Transmission is  added url-list options  start of .torrent file like this.

![image](https://user-images.githubusercontent.com/36080488/94421750-f888de00-01c0-11eb-9865-7b809167e003.png)

Probably, it seems to be a problem due to the different method of writing .torrent for each torrent client.

[ from google translate english  lol  :) ]

parsing error here.

![image](https://user-images.githubusercontent.com/36080488/94422175-a1cfd400-01c1-11eb-9808-f4d34e1a64a4.png)

MonoTorrent.BEncoding.BEncodeDecoder.cs
![image](https://user-images.githubusercontent.com/36080488/94422309-df346180-01c1-11eb-99d1-5fa63b634f23.png)

samhocevar:
This is Tixati not fully following the torrent specification. It must store keys in lexicographical order, but as you can see the “url-list” entry is stored before the “created by” entry. Some torrent clients are more liberal than others in what they accept (I think MonoTorrent could accept to load such torrents, what do you think @alanmcgovern?)

I have posted a bug report on the Tixati website linking to this issue.

alanmcgovern:
@samhocevar @wizard872 

The biggest issue with these kinds of torrents is that some clients will parse them, put the keys in the correct order, then compute the infohash. Other clients will compute the infohash using keys in the incorrect order.

[...]

PR #341, opened 2020-11-24
[core] Compute the correct infohash for misordered dictionaries

alanmcgovern:
If a BEncodedDictionary has been encoded with the keys in the
wrong order, we will now compute the infohash based on the
underlying bytes. This is the other spec compliant approach.

The first spec compliant approach is to reject the torrent.
Unfortunately there are plenty of invalid torrents out there,
so at this stage we may as well do the work to support this.

This should not be any worse performance wise as the original
approach.

Solves https://github.com/alanmcgovern/monotorrent/issues/332


alanmcgovern:
Turns out I had most of the code in place already to make this really really easy :)

Implemented!

PR #346, opened 2020-11-29
two bugs fixed

OneFingerCodingWarrior:


src/MonoTorrent/MonoTorrent.Client/Managers/InactivePeerManager.cs
```diff
@@ -103,6 +108,11 @@ internal void TimePassed ()
                         }
                     }
                 }
+                }
+                catch (Exception ex)
+                {
+                    Debug.WriteLine("TimePassed Error " + ex.Message);
```
vlasenkoalexey, line 114: Instead of doing Debug.WriteLine here and in other places, could you create a class logger as static readonly Logger logger = Logger.Create (); and then log it as logger.Debug(...

src/MonoTorrent/MonoTorrent.Client/Managers/InactivePeerManager.cs
```diff
@@ -61,8 +62,12 @@ internal void TimePassed ()
             int candidateSecondsConnected ;
             int candidateSecondsSinceLastBlock;
             int candidateDataBytes;
-            for (int i = 0; i < TorrentManager.Peers.ConnectedPeers.Count; i++) {
-                PeerId nextPeer = TorrentManager.Peers.ConnectedPeers[i];
+            PeerId[] arr = TorrentManager.Peers.ConnectedPeers.ToArray();
```
vlasenkoalexey, line 65: what's the point in converting to array or to list here and below?

OneFingerCodingWarrior: if function CleanupSocket() changed the ConnectedPeers, exception argument out of range might get throwed. i'm not sure this is a good fix or not

src/MonoTorrent/MonoTorrent.Client.Tracker/TrackerTier.cs
```diff
@@ -216,6 +216,8 @@ internal async ReusableTask ScrapeAsync (ScrapeParameters args, CancellationToke
 
             for (int i = 0; i < Trackers.Count; i++) {
                 var tracker = Trackers[(ActiveTrackerIndex + i) % Trackers.Count];
+                if (false == tracker.CanScrape)
```
alanmcgovern, line 219:
```suggestion
                if (!tracker.CanScrape)
```

It's a bit more idiomatic to write it this way :)

src/MonoTorrent/MonoTorrent.Client/Managers/InactivePeerManager.cs
alanmcgovern: CleanupSocket is guaranteed to run on the `ClientEngine.MainLoop` thread. This code is also guaranteed to run on that thread. As such it should be impossible for CleanupSocket to run while this code is running, unless this code calls CleanupSocket.

I think we can undo this part of the patch based on that. Right?

src/MonoTorrent/MonoTorrent.Client/Managers/InactivePeerManager.cs
alanmcgovern: Definitely +1 on the Logger.Create change :)

src/MonoTorrent/MonoTorrent.Client/Managers/InactivePeerManager.cs
```diff
@@ -121,8 +131,8 @@ internal void TimePassed ()
 
             // We've found a peer to disconnect
             // Add it to the inactive list for this torrent and disconnect it
-            InactivePeerList.Add (TorrentManager.Peers.ConnectedPeers[peerToDisconnect].Uri);
-            ConnectionManager.CleanupSocket (TorrentManager, TorrentManager.Peers.ConnectedPeers[peerToDisconnect]);
```
alanmcgovern, line 125: Just to make sure my understanding is correct - this is executing *after* we've finished looping, so we're not mutating the `ConnectedPeers` list while we're iterating over it. Right?

src/MonoTorrent/MonoTorrent.Client/Modes/DownloadMode.cs
```diff
@@ -65,14 +66,25 @@ public override void Tick (int counter)
...
+                    {
+                        ConnectionManager.CleanupSocket(Manager, arrPID[i]);
+                    }
+                }
+                catch (System.Exception ex)
+                {
+                    Debug.WriteLine("Tick Error " + ex.Message);
                 }
             }
         }
```
alanmcgovern, line 86: I think the old and new code are functionally the same, except CleanupSocket has been wrapped in a try/catch. Were you seeing exceptions being raised in this method? the expectation for MonoTorrent is that 'CleanupSocket' never raises an error.

src/MonoTorrent/MonoTorrent.Client/Modes/InitialSeedingMode.cs
```diff
@@ -92,7 +92,7 @@ public override void Tick (int counter)
             if (unchoker.Complete) {
                 PeerMessage bitfieldMessage = new BitfieldMessage (Manager.Bitfield);
                 PeerMessage haveAllMessage = new HaveAllMessage ();
-                foreach (PeerId peer in Manager.Peers.ConnectedPeers) {
+                foreach (PeerId peer in Manager.Peers.ConnectedPeers.ToArray()) {
```
alanmcgovern, line 95: Same here - ConnectedPeers cannot be mutated while this code is running (if it were possible then it could also mutate while `ToArray` is being run, which could result in similar issues).

src/MonoTorrent/MonoTorrent.Client/Modes/MetadataMode.cs
```diff
@@ -189,7 +189,15 @@ protected override void HandleLtMetadataMessage (PeerId id, LTMetadata message)
...
                                         Directory.CreateDirectory (Path.GetDirectoryName (savePath));
-                                    File.Delete (savePath);
+
+                                    if(true == File.Exists(savePath))
+                                        File.Delete (savePath);
+                                    else//savePath is a directory
+                                    {
+                                        savePath += "\\" + t.Name + ".torrent";
+                                        using var fs = File.Create(savePath);
+                                    }
```
alanmcgovern, line 199: I'd like to do a more holistic refactor so that all string parameters passed to the library are renamed so it's super clear whether they refer to a `directory`or a `file` At the moment it's pretty unclear whether a given string is supposed to be an actual file path or a directory name.  It's also unclear whether something is a full path or relative path.

It shouldn't be this confusing!

I love the direction this change is moving. I think with a little more planning we can perfect this!

src/MonoTorrent/MonoTorrent.Client/Modes/MetadataMode.cs
```diff
@@ -189,7 +189,15 @@ protected override void HandleLtMetadataMessage (PeerId id, LTMetadata message)
                                 try {
                                     if (!Directory.Exists (Path.GetDirectoryName (savePath)))
                                         Directory.CreateDirectory (Path.GetDirectoryName (savePath));
-                                    File.Delete (savePath);
+
+                                    if(true == File.Exists(savePath))
+                                        File.Delete (savePath);
+                                    else//savePath is a directory
+                                    {
+                                        savePath += "\\" + t.Name + ".torrent";
```
alanmcgovern, line 197: We should use `Path.Combine` so paths are combined correctly on Windows and Linux/MacOS

src/MonoTorrent/MonoTorrent.Client/Modes/Mode.cs
```diff
@@ -85,50 +86,50 @@ public void HandleMessage (PeerId id, PeerMessage message)
...
+            else if (message is LTMetadata metadata)
+                HandleLtMetadataMessage (id, metadata);
+            else if (message is LTChat chat)
+                HandleLtChat (id, chat);
+            else if (message is PeerExchangeMessage peerexchange)
+                HandlePeerExchangeMessage (id, peerexchange);
+            else if (message is HandshakeMessage handshake)
+                HandleHandshakeMessage (id, handshake);
+            else if (message is ExtensionMessage extension)
+                HandleGenericExtensionMessage (id, extension);
```
alanmcgovern, line 132: Can you submit this as a separate PR. It's a very self-contained change and could merge all by itself :)

+1 on using the modern syntax here!

src/MonoTorrent/MonoTorrent.Client/Modes/Mode.cs
```diff
@@ -532,9 +531,11 @@ void PreLogicTick (int counter)
             if (Manager.finishedPieces.Count > 0)
                 SendHaveMessagesToAll ();
 
-            for (int i = 0; i < Manager.Peers.ConnectedPeers.Count; i++) {
-                id = Manager.Peers.ConnectedPeers[i];
-                if (id.Connection == null)
+            foreach (PeerId id in Manager.Peers.ConnectedPeers.ToArray())
```
alanmcgovern, line 534: I don't want to repeat too much, but the `ToArray` shouldn't be necessary anywhere and could result in a fairly high number of allocations :)

src/MonoTorrent/MonoTorrent.Client/Modes/MetadataMode.cs
alanmcgovern: I don't want to do this larger change *right now* as i think there's value in merging your patch today without trying to fix up the rest of the library. I think we should document this value as being a *directory*, and internally let's always treat it as a directory. Keeping the `if (File.Exists(path)) File.delete(path)` check should be safe as it ensures we can always create a directory in that path to store our metadata in.

Could you submit this section in it's own PR so I can review/test it before merging? I'd love to test these scenarios to see what happens. I think the first scenario will still cause an issue with this modified code.

1 - pass `save\with\sub-dir` as the metadata save path. Do not create any of those directories to begin with.
2 - pass `save` as the metadata safe path. Do not create the `save` directory first

OneFingerCodingWarrior:
ok, i will submit a separate PR

PR #352, opened 2020-12-09
Enable memory cache by default

alanmcgovern:


alanmcgovern:
When combined with a patch to disable the 'ContinueAnyExistingRequest' logic in the piece picker, monotorrent can download a 2.8GB torrent with these disk read/write stats:
```
Total Read:         62,816.00 kB
Total Written:      2,871,610.00 kB
```
As you can see the vast majority of the torrent could be successfully downloaded and hashed without reading data back from disk. A handful of pieces arrived in an incorrect order (for whatever reason) and weren't retained by the 5MB cache long enough to hit the fast path.

Still, this is pretty good!

Unfortunately if you 'ContinueAnyExistingRequest' logic is left in place, too many blocks are received out of order (as they are being requested from multiple peers) and we end up re-reading hundreds of MB from disk. I'd like to see if we can finesse the 'ContinueAnyExistingRequest' logic before merging this, though it is a reasonable improvement in and of itself.

alanmcgovern:
Two small tweaks to `ContinueAnyExistingPiece` have made this much less of an issue! https://github.com/alanmcgovern/monotorrent/pull/353

The stats are more like this now:

```
Total Read:            125,318.00 kB
Total Written:      2,871,610.00 kB
```
125MB of misordered blocks out of 2.8GB is pretty alright. Before this patch it was usually over 700MB and frequently over 1.5GB.

PR #369, opened 2020-12-19
[core] Allow prioritisation of encryption methods

alanmcgovern:
The encryption spec does not allow the initiator of the
connection to set a preference. As such, when we want
to establish a connection with a remote peer we should
only offer 1 encryption method (RC4 Header *or* RC4Full)
to force them to choose our preferred method.

If the connection is rejected, we'll try again with the
other one (if possible).


alanmcgovern:
Not sure why the build isn't reporting a status on this PR. Perhaps i exceeded a rate limit unexpectedly? The build is green anyway :)

alanmcgovern:
Aaand there we go. The build status showed up eventually!

PR #385, opened 2021-01-06
Added CreateSubFolder property to TorrentSetting

AN3Orik:
Added CreateSubFolder property to TorrentSetting (if set to false root folder will be not created for multi-file torrent)
Usefull when MonoTorrent using in application as self-update engine.

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
@@ -104,7 +104,7 @@ namespace MonoTorrent.Client
         /// <summary>
         /// If set to false then root folder will be not created for multi-files torrents
         /// </summary>
-        internal bool CreateSubFolder { get; } = true;
+        public bool CreateSubFolder { get; } = true;
 
         public TorrentSettings ()
         {
--- a/src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
+++ b/src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
@@ -101,12 +101,17 @@ namespace MonoTorrent.Client
         /// </summary>
         internal TimeSpan TimeToWaitUntilIdle => TimeSpan.FromMinutes (10);
 
+        /// <summary>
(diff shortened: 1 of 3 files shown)


alanmcgovern:
This is a great idea! I love the simplicity this offers to allow people to choose between the two most common options: Create a containing directory when it's a multi-file torrent, or do not create one.

Technically this is configurable already by using the second `string` parameter when instantiating a new TorrentManager. However this is not obvious and is trickier to use. 

I will refactor the constructor for TorrentManager to remove this extra complication, and will merge this patch in. If someone has complex requirements on *exactly* where files should be placed they can use the `DiskManager.MoveAsync` API to move the files to different locations. For example this could be used if you wanted something more complicated than "Create a sub directory: yes or no".

src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettings.cs
```diff
@@ -101,12 +101,17 @@ public class TorrentSettings : IEquatable<TorrentSettings>
         /// </summary>
         internal TimeSpan TimeToWaitUntilIdle => TimeSpan.FromMinutes (10);
 
+        /// <summary>
+        /// If set to false then root folder will be not created for multi-files torrents
+        /// </summary>
+        internal bool CreateSubFolder { get; } = true;
```
alanmcgovern, line 107:
```suggestion
        public bool CreateSubFolder { get; } = true;
```

src/MonoTorrent/MonoTorrent.Client/Settings/TorrentSettingsBuilder.cs
```diff
@@ -99,6 +99,11 @@ public class TorrentSettingsBuilder
         /// </summary>
         public int WebSeedSpeedTrigger { get; set; }
 
+        /// <summary>
+        /// If set to false then root folder will be not created for multi-files torrents
```
alanmcgovern, line 103:
```suggestion
        /// If set to false then root folder will be not created for multi-files torrents, otherwise all files will be placed inside a root folder named <see cref="TorrentManager.Name"/>. Defaults to true.
```

alanmcgovern (approved):
Two small tweaks! :D

PR #386, opened 2021-01-06
EncryptionType allowedEncryption AIOOB fix

AN3Orik:
Fixed ArrayIndexOutOfBound exception when peer allowedEncryption list is empty


alanmcgovern:
Great! Thanks for the patch for this. I've merged it and added a test covering the behaviour!

PR #387, opened 2021-01-06
EncryptionType ArrayIndexOutOfBound exception fix

AN3Orik:
Fixed ArrayIndexOutOfBound exception when peer allowedEncryption list is empty

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Client.Encryption/EncryptionType.cs
+++ b/src/MonoTorrent/MonoTorrent.Client.Encryption/EncryptionType.cs
@@ -54,12 +54,14 @@ namespace MonoTorrent.Client
 
         internal static IList<EncryptionType> Remove (IList<EncryptionType> allowedEncryption, EncryptionType encryptionType)
         {
-            var result = new EncryptionType[allowedEncryption.Count - 1];
-            int j = 0;
-            for (int i = 0; i < allowedEncryption.Count; i++)
-                if (allowedEncryption[i] != encryptionType)
-                    result[j++] = allowedEncryption[i];
-
+            var result = allowedEncryption;
+            if (result.Count > 0) {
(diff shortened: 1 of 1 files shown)


alanmcgovern:
Great stuff! I'll merge this and add a test in a separate commit!

src/MonoTorrent/MonoTorrent.Client.Encryption/EncryptionType.cs
```diff
@@ -54,12 +54,14 @@ internal static bool SupportsRC4 (IList<EncryptionType> allowedEncryption)
...
         internal static IList<EncryptionType> Remove (IList<EncryptionType> allowedEncryption, EncryptionType encryptionType)
         {
-            var result = new EncryptionType[allowedEncryption.Count - 1];
-            int j = 0;
-            for (int i = 0; i < allowedEncryption.Count; i++)
-                if (allowedEncryption[i] != encryptionType)
-                    result[j++] = allowedEncryption[i];
-
+            var result = allowedEncryption;
+            if (result.Count > 0) {
```
alanmcgovern, line 58: While writing a test I realised that this introduced an unexpected change.

```
var result = EncryptionTypes.Remove (new List<EncryptionType> (), EncryptionType.PlainText);
result.Add (EncryptionType.PlainText);
```

This should throw an exception as `result` should be immutable. With your initial change this was no longer true. I tweaked it to always return an immutable empty collection when the input list is empty!

alanmcgovern: Oh derp, my comment is incorrect - the original code was fine in that respect. It would still have been wrapped with the `MakeReadOnly` at the end of the call. Never mind! That's my bad for mis-reading after the original review 😆

PR #403, opened 2021-02-19
Torrent SavePath fix

justscribe:
I'd say it is incorrect behavior, when TorrentManager automatically chosing the way how files will be saved. When I'm set a SavePath, I want to be confident that files will be saved there, without any additional ierarchy. In my case, such behavior makes impossible to save/update the folder with new files if Torrent or SavePath name will be changed. So, I've added an additional config value, to create or not a subdirectory for Torrent.


justscribe:
Oops, duplicated... So not only I need this)

alanmcgovern:
Sorry for not noticing the duplication! Thanks for the patch, you were just a little unlucky in that I saw the other one first! It was a great suggestion :)

alanmcgovern:
Also - I think I want to rename it to SubDirectory instead of SubFolder. Its a bit more idiomatic. Your patch definitely got that aspect a little bit more-correct :)

justscribe:
No problems, great, that we have such config in any form)

PR #408, opened 2021-03-06
Engine addremove

alanmcgovern:
Some simplifications to the API.

Rather than manually creating a torrent manager, then registering it, we can make this a single API call: `ClientEngine.AddAsync`

Similarly, rather than creating a StreamingProvider to set up a torrent for streaming access, you can now just call `ClientEngine.AddStreamingAsync`. This will populate the `TorrentManager.StreamProvider` property and you'll be able to get access to the files within the torrent while the torrent is actively downloading. 

As always, if you use `TorrentManager.StreamProvider` to create a stream you can treat it like a normal FileStream. If you call ReadAsync, SeekAsync, etc, the engine will automatically download the data required to fulfill your Read or Seek request.

### DIFF
--- a/src/MonoTorrent.Tests/Client/ClientEngineTests.cs
+++ b/src/MonoTorrent.Tests/Client/ClientEngineTests.cs
@@ -91,8 +91,7 @@ namespace MonoTorrent.Client
                 Private = true
             };
 
-            var manager = new TorrentManager (editor.ToTorrent (), "path", new TorrentSettings ());
-            await rig.Engine.Register (manager);
+            var manager = await rig.Engine.AddAsync (editor.ToTorrent (), "path", new TorrentSettings ());
 
             var dht = (ManualDhtEngine) rig.Engine.DhtEngine;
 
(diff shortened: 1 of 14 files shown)


OneFingerCodingWarrior:
is it safe to call Engine.AddAsync before manager.LoadFastResume (new FastResume ((BEncodedDictionary) fastResume[torrent.InfoHash.ToHex ()]));? I noticed State == TorrentState.Stopped if required for LoadFastResume

OneFingerCodingWarrior:
is it safe to call Engine.AddAsync before manager.LoadFastResume (new FastResume ((BEncodedDictionary) fastResume[torrent.InfoHash.ToHex ()]));? I noticed State == TorrentState.Stopped is required for LoadFastResume

alanmcgovern:
Yes it is!

You must call this to create a TorrentManager now. AddAsync is the same as creating a TorrentManager manually and then calling ClientEngine.Register.

You can call LoadFastResume after the manager has been registered if you are still using the old way of doing things.

PR #415, opened 2021-03-20
[core] Add a fast path when no files exist

alanmcgovern:
This can be a bit faster than binary searching over all files
every time we emit a fake 'piece hashed' event indicating
data which *does not exist* has 'failed' a hash check.


alanmcgovern:
Let's skip a fast path. It's only a tiny bit noticeable for torrents with 32k pieces.

Issue #418, opened 2021-03-21
Make IBitfield read-only wrapper


alanmcgovern:
It's easy to corrupt state because this is public read/write

alanmcgovern:
Solved

PR #422, opened 2021-03-23
[core] Make the 'BitField' mutating methods internal

alanmcgovern:
The 'BitField' class is supposed to be readonly in every scenario
where it's exposed to users.

After a fair bit of thinking, the best way to enforce this is to
make the mutating methods internal. This will allow the monotorrent
to update the bitfields as data is modified, while ensuring users
cannot accidentally corrupt data by using the mutating methods
unexpectedly.

If, for some reason, users of monotorrent need a mutable bitfield
then that can be exposed. However a better choice may be for them
to use the built in System.Collections.BitArray class.

### DIFF
--- a/src/MonoTorrent/MonoTorrent/BitField.cs
+++ b/src/MonoTorrent/MonoTorrent/BitField.cs
@@ -47,13 +47,17 @@ namespace MonoTorrent
 
         readonly int[] array;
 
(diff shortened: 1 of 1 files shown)


alanmcgovern:
Addresses https://github.com/alanmcgovern/monotorrent/issues/418

PR #432, opened 2021-04-06
fast resume file modification

OneFingerCodingWarrior:


src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
```diff
@@ -1001,7 +1001,7 @@ internal async ReusableTask MaybeLoadFastResumeAsync ()
             }
         }
 
-        internal async ReusableTask MaybeWriteFastResumeAsync ()
+        public async ReusableTask MaybeWriteFastResumeAsync ()
```
alanmcgovern, line 1004: > change MaybeWriteFastResumeAsync to public for Saving Fast Resume file periodic,or before app shutdown due to exception

Perhaps this could be best supported as a setting in `EngineSettings` ? `public enum FastResumeMode { Lossless, Lossy }`.

In lossless mode you are guaranteed that you will never have to redownload data which has passed it's hash check as the fast resume data will only be deleted from disk when a torrent enters 'Downloading' mode and will only be written to disk when the torrent stops, or enters 'seeding' mode. In the event of a crash while downloading you have to hash check the entire file before downloading again.

In 'lossy' mode fast resume data will be stored regularly on disk, but in the event of a crash you will lose the most recent N seconds/minutes of downloaded data as the fast resume file will be slightly stale.

src/MonoTorrent/MonoTorrent.Client/ClientEngine.cs
```diff
@@ -308,6 +308,7 @@ public async Task<bool> RemoveAsync (TorrentManager manager)
             manager.Engine = null;
             manager.DownloadLimiters.Remove (downloadLimiters);
             manager.UploadLimiters.Remove (uploadLimiters);
+            _ = manager.MaybeDeleteFastResumeAsync ();
```
alanmcgovern, line 311: I had a task, which i forgot about, to address the issue of stale data.

I was wondering if this pair of methods would help:

```
// Iterate over all cache files. Delete any file which *does not* correspond to a torrent loaded in the engine.
ClientEngine.ClearStaleCacheFiles ();

// Deletes *all* cached fast resume data, dht nodes, etc etc etc
ClientEngine.ClearCacheFiles ();

// Deletes all fast resume data, or other torrent-specific data, for the supplied TorrentManager/MagnetLink/InfoHash
ClientEngine.ClearCacheFiles (InfoHash|TorrentManager|MagnetLink)
``` 

Usecases include:
```
async Task RemoveFromEngine (TorrentManager manager)
{
    // from the torrent from the engine
    await engine.RemoveAsync (manager);

    // Remove any cache files for torrents which are no longer in the engine
    await engine.ClearStaleCacheFiles ();

    // Or just remove the data for that torrent specifically:
    await engine.ClearCacheFiles (manager);
}
```

If something goes wrong at some point and you want to clear data before loading a torrent...
```
async Task AddToEngine (Torrent torrent)
{
    await engine.ClearCacheFiles (torrent.InfoHash);
    await engine.AddAsync (torrent);
}
``` 

Maybe there could be an enum too which lists the kinds of data which will be cleared, so you could selectively clear fast resume, or something else?

OneFingerCodingWarrior:
i agree with you, to encapsulate fast resume logic in ClientEngine is better than call the related API explicitly

src/MonoTorrent/MonoTorrent.Client/Managers/TorrentManager.cs
OneFingerCodingWarrior: how about use Lossy as default value ? usually files are very big, and even worse if there are many files ,

alanmcgovern: https://github.com/alanmcgovern/monotorrent/tree/fast-resume-improvements

PR #454, opened 2021-09-02
Split library

alanmcgovern:
I thought it would be nicer to split this into a few separate assemblies.

It's a bit easier to reason about the piece picking code, or other key aspects, when the referenced assemblies/files is a smaller set. It'll have no real impact on users of the library as they'll just have a few extra assembly references in the end.


OneFingerCodingWarrior:
Wow, that's a huge refactor

PR #481, opened 2021-11-23
example bitorrent v2 initial parsing

alanmcgovern:


alanmcgovern:
https://blog.libtorrent.org/2020/09/bittorrent-v2/

Tested using the v2 torrent linked there. 

Ideally the hybrid torrent would be used to validate support. Monotorrent should be able to load that and use the v1 metadata to download, and then (eventually) transition to using the v2 metadata

PR #490, opened 2021-12-12
[dht] Use SpeedMonitor to calculate DHT data rates

alanmcgovern:
Record total bytes sent/received, as well as the estimated transfer rate in bytes/second.

### DIFF
--- a/src/MonoTorrent/MonoTorrent.Dht/IDhtEngine.cs
+++ b/src/MonoTorrent/MonoTorrent.Dht/IDhtEngine.cs
@@ -35,6 +35,29 @@ using MonoTorrent.Connections.Dht;
 
 namespace MonoTorrent.Dht
 {
+    public interface ITransferMonitor
+    {
+        /// <summary>
+        /// Total bytes sent since the start of the session.
+        /// </summary>
+        long BytesSent { get; }
+
+        /// <summary>
+        /// Total bytes received since the start of the session.
+        /// </summary>
+        long BytesReceived { get; }
+
+        /// <summary>
+        /// Estimate of the amount of data sent every second, in bytes/second.
+        /// </summary>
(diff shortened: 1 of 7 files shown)


alanmcgovern:
@YegorStepanov 

This should give you the DHT stats you needed. Let me know what you think!

src/MonoTorrent/MonoTorrent.Dht/IDhtEngine.cs
```diff
@@ -35,13 +35,37 @@
...
+
+        /// <summary>
+        /// Estimate of the amount of data sent every second, in bytes/second.
+        /// </summary>
+        long ReceiveRate { get; }
+
+        /// <summary>
+        /// Estimate of the amount of data sent every second, in bytes/second.
+        /// </summary>
+        long SendRate { get; }
```
YegorStepanov, line 58: I'm not sure how useful `SendRate/ReceiveRate`. If I grab the value every second, I want to know the average of the one second. Idk when I want to use the average of the 12 seconds (12 is hard-coded), so it's 'some average'. This value can be displayed only. It cannot be used for any kind of processing.

YegorStepanov (commented):
Thanks

src/MonoTorrent/MonoTorrent.Dht/IDhtEngine.cs
alanmcgovern: In that case I suspect the best solution for you is to iterate over all active connections every N seconds, grab the total bytes sent/received and calculate the delta since the previous time your logic ran. 

It would be very hard to provide anything else which isn't prone to skipping a value, or duplicating a value. 

If the engine calculated the "average" over the last second, it would be hard to ensure you (the consumer of the library) always read the number *exactly* once per second.

Issue #491, opened 2021-12-15
Invalid Characters in Path


Niegen:
Hello,

When attempting to run the following:

`await Engine.AddAsync(torrent, outputDirectory);`

on a torrent whose name contains double quotes "

the following gets thrown:

```
Exception thrown: 'System.ArgumentException' in mscorlib.dll
Exception thrown: 'System.AggregateException' in mscorlib.dll
System.AggregateException: Une ou plusieurs erreurs se sont produites. ---> System.ArgumentException: **Caractères non conformes dans le chemin d'accès**.
   à System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   à System.IO.Path.Combine(String path1, String path2)
   à MonoTorrent.Client.TorrentManager.SetMetadata(Torrent torrent)
   à MonoTorrent.Client.ClientEngine.<AddAsync>d__88.MoveNext()
--- Fin de la trace de la pile à partir de l'emplacement précédent au niveau duquel l'exception a été levée ---
   à System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   à System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   à MonoTorrent.Client.ClientEngine.<AddAsync>d__87.MoveNext()
--- Fin de la trace de la pile à partir de l'emplacement précédent au niveau duquel l'exception a été levée ---
   à System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   à System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   à System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   --- Fin de la trace de la pile d'exception interne ---
   à System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   à System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   à System.Threading.Tasks.Task.Wait()
---> (Exception interne #0) System.ArgumentException: Caractères non conformes dans le chemin d'accès.
   à System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   à System.IO.Path.Combine(String path1, String path2)
   à MonoTorrent.Client.TorrentManager.SetMetadata(Torrent torrent)
[...]

PR #493, opened 2021-12-21
[core] Replace invalid characters in the generated path

alanmcgovern:
Replace invalid characters with the hex equivalent (and hope
there is no collision).

Addresses https://github.com/alanmcgovern/monotorrent/issues/491


Niegen:
Hi Alan,

Thanks for the fix.

The torrent downloaded correctly, however, the " were replaced by the string **22**.

Not a huge problem by itself, but then I cannot locate the actual final torrent folder by doing 

`Path.Combine(Options.Instance.OutputDirectory, torrent.Name)`

since torrent.Name doesn't reflect the "fixed" folder name and still contains the invalid characters.

Maybe you should add a read only property for the final folder full path?

You are very close though. Thanks again.

alanmcgovern:
You can find the full paths to all files by accessing the TorrentManager.Files list and checking the FullPath property. 

In the meantime, ill have a think about the best way to expose a "ContainingDirectory" property which represents the sanitized path!

alanmcgovern:
There's a new `TorrentManager.ContainingDirectory` property. I added xml docs which explain the difference between this and `TorrentManager.SavePath`. For a single file torrent `ContainingDirectory` and `SavePath` are the same. For a multi-file torrent, when `TorrentSettings.CreateContainingDirectory` is enabled, the escaped version of `Torrent.Name` is appended to `SavePath` and is stored in the `ContainingDirectory` property.

If you use the public APIs to move files to different locations then these properties will have no clear/accurate meaning. However `TorrentFileInfo.FullPath`  will always be correct. You can get this by looping over all the files in a torrentmanager :) 

I'll merge this, and release an updated nuget package, as it seems to solve your first issue, and your followup issue has been solved too!

Do let me know if you have any further issues/questions!

PR #498, opened 2022-01-06
[core] Allow to load FastResume objects from a stream

samhocevar:
Since `FastResume.Encode()` can write to a stream, it makes sense to allow `FastResume.TryLoad()` to read from a stream. This allows a client application to manage collections of `FastResume` objects.

Right now there is no other way because all constructors are internal.


alanmcgovern (approved):
Looks great to me!

alanmcgovern:
Just in case you didn't know: You don't need to save fast resume data for individual torrents - you can save/restore the entire ClientEngine! This is a single method invocation and saves/restores EngineSettings, TorrentSettings, FastResume, if you've moved any TorrentFiles to non-default locations that is saved/restored too.

Issue #517, opened 2022-03-20
Monotorrent fails to gracefully handle absolute paths in torrent files


issueg2k4g34j2g:
## Expected Behavior
monotorrent should silently rewrite absolute paths into relative paths instead of throwing an error message.

## Current Behavior
LoadTorrentFiles chokes when supplied with a torrent file containing absolute paths.

The responsible code can be found at https://github.com/alanmcgovern/monotorrent/blob/999d34b5f102aa4dfb72497b381e8f36b86f5653/src/MonoTorrent.Client/MonoTorrent/PathValidator.cs#L38-L46

and

https://github.com/alanmcgovern/monotorrent/blob/b0c484684c708bfee4d92c06178f1e64d2cc4399/src/MonoTorrent.Client/MonoTorrent/Torrent.cs#L646-L655

## Possible Solution
All common torrent clients deal with this gracefully by rewriting absolute paths into relative paths. 

A simple solution would be to simply disregard leading empty path segments (and slashes?) while parsing the paths in LoadTorrentFilesV1. 

This would also be a good solution from a security POV, as it would not require modifications to the path validation code.

## Steps to Reproduce
1.  Attempt to load this torrent file with monotorrent
[Fargo.S01.1080p.BluRay.DTS5.1.x264-ROVERS.[BTN].torrent.zip](https://github.com/alanmcgovern/monotorrent/files/8311854/Fargo.S01.1080p.BluRay.DTS5.1.x264-ROVERS.BTN.torrent.zip)

## Context (Environment)
I have run into several torrent files demonstrating this issue while diagnosing issues with my sonarr setup, they appear to be fairly common in the wild.
Most of these files appear to originate from ruTorrent, which seems to be a fairly common software used to produce torrent files. Other software (deluge, transmission, rtorrent, qbittorrent, etc...) handle these files gracefully, so they are not treated as corrupt by various indexers.

These files appear to be fully spec compliant, just strange.

issueg2k4g34j2g:
My apologies for not including a pull request despite this being such a simple issue, I'm completely out of my depth with C#.

issueg2k4g34j2g:
So I figured I'd try to make a useful pull request anyway, this *should work* as far as I can tell.

It is worth considering whether or not this should be only limited to leading directory separators, but I can't imagine a legitimate use case for having extra directory separators anywhere in the path.

alanmcgovern:
Your analysis was *near* perfect!

[...]

PR #518, opened 2022-03-21
Disregard certain path segments

issueg2k4g34j2g:
This *should* solve #517, but I don't know C# 🙃

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent/Torrent.cs
+++ b/src/MonoTorrent.Client/MonoTorrent/Torrent.cs
@@ -636,8 +636,10 @@ namespace MonoTorrent
 
                         case ("path.utf-8"):
                             foreach (BEncodedString str in ((BEncodedList) keypair.Value)) {
-                                sb.Append (str.Text);
-                                sb.Append (Path.DirectorySeparatorChar);
+                                if (!(string.IsNullOrEmpty (str.Text) || str.Text == Path.DirectorySeparatorChar)) {
+                                    sb.Append (str.Text);
+                                    sb.Append (Path.DirectorySeparatorChar);
+                                }
                             }
                             path = sb.ToString (0, sb.Length - 1);
                             sb.Remove (0, sb.Length);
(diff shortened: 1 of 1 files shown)


src/MonoTorrent.Client/MonoTorrent/Torrent.cs
```diff
@@ -636,8 +636,10 @@ static IList<ITorrentFile> LoadTorrentFilesV1 (BEncodedList list, int pieceLengt
 
                         case ("path.utf-8"):
                             foreach (BEncodedString str in ((BEncodedList) keypair.Value)) {
-                                sb.Append (str.Text);
-                                sb.Append (Path.DirectorySeparatorChar);
+                                if (!(string.IsNullOrEmpty (str.Text) || str.Text == Path.DirectorySeparatorChar)) {
```
alanmcgovern, line 639:
```suggestion
                                if (!BEncodedString.IsNullOrEmpty (str)) {
```

src/MonoTorrent.Client/MonoTorrent/Torrent.cs
```diff
@@ -646,8 +648,10 @@ static IList<ITorrentFile> LoadTorrentFilesV1 (BEncodedList list, int pieceLengt
                         case ("path"):
                             if (string.IsNullOrEmpty (path)) {
                                 foreach (BEncodedString str in ((BEncodedList) keypair.Value)) {
-                                    sb.Append (str.Text);
-                                    sb.Append (Path.DirectorySeparatorChar);
+                                    if (!(string.IsNullOrEmpty (str.Text) || str.Text == Path.DirectorySeparatorChar)) {
```
alanmcgovern, line 651:
```suggestion
                                    if (!BEncodedString.IsNullOrEmpty (str)) {
```

alanmcgovern:
Thanks for the patch! I tweaked it, and added a test in a separate follow up commit. I'll also backport it as it's an easy fix!

PR #526, opened 2022-04-12
bep47

jpmikkers:
Todo:
- partition function tests
- md5 hashing tests, should not be affected by padding
- ~~not sure what to do on torrent.cs line 429 for hybrid torrents..~~


src/MonoTorrent.Client/MonoTorrent/Torrent.cs
```diff
@@ -426,7 +426,7 @@ void ProcessInfo (BEncodedDictionary dictionary, ref PieceHashesV1? hashesV1)
                     case ("files"):
                         // This is the list of files using the v1 torrent format.
                         // Only load if we have not processed filesv2
-                        if (Files.Count == 0)
+                        // if (Files.Count == 0) // TODO JMIK: merge v1 hashes info with v2 ??
```
alanmcgovern, line 429: I'm lazy, so I use https://chocobo1.github.io/bencode_online for viewing the content of .torrent files when I want to review exactly what's in there.

I used that site to view the hybrid torrent linked in the libtorrent blog ( https://blog.libtorrent.org/2020/09/bittorrent-v2/ ) to see what impact choosing the V1 or V2 files would have.

The difference is:
- Padding files only exist in V1 metadata
- `pieces root` only exists with V2 metadata.

In order to support hybrid torrents, there is one way I can think of to avoid merging data, and it also provides a little bit of extra safety for incorrectly generated torrents.

For a standards compliant hybrid torrent we know padding files are always exactly: `TorrentManager.PieceLength - file.Length % TorrentManager.PieceLength` bytes long. (They always extend the file so it comes the remaining space in that piece).

As such, when we call [LoadTorrentFilesV2](https://github.com/alanmcgovern/monotorrent/blob/002708a4220eb4c2d2b6d950c35d5bdae7712184/src/MonoTorrent.Client/MonoTorrent/Torrent.cs#L435) we could pass in an additional `isHybrid` boolean (i.e. `hasV1Data  && hasV2Data` ) and set the correct padding value for the files we parse if the torrent is a hybrid v1/v2 torrent.

Later, when we parse the V1 metadata we can verify that the *actual* padding length parsed from the v1 'files' key matches the value we generated when parsing the v2 files from `files tree`. If it doesn't match then it's not a standards compliant file and we can abort parsing.

I think that solves the problem, while also adding a sanity check that the padding files were created correctly :)

alanmcgovern:
I didn't review it in full - i'll do that tomorrow!

src/MonoTorrent.Client/MonoTorrent/Torrent.cs
jpmikkers: Added that in my latest commit, the validation works for the hybrid test torrent. The hybrid torrent loading test fails with MonoTorrent.TorrentException : the 'piece layers' dictionary did not contain an entry for the file 'tbl-tint.mpg'.
I think I spotted a rounding bug in v2 EndPieceIndex calculation..

src/MonoTorrent.Client/MonoTorrent/TorrentCreator.cs
```diff
@@ -516,6 +537,25 @@ static BEncodedValue ToFileInfoDict (InputFile file)
...
+            var filePath = new BEncodedList ();
+            filePath.Add (new BEncodedString (".pad"));
+            filePath.Add (new BEncodedString ($"{file.Padding}"));
+
+            fileDict["length"] = new BEncodedNumber (file.Padding);
+            fileDict["path"] = filePath;
+
+            // TODO JMIK: do padding files have MD5 hashes? It's easy to add an all-zero MD5 hash calculation here
+            //if (file.MD5 != null)
+            //    fileDict["md5sum"] = (BEncodedString) file.MD5;
```
alanmcgovern, line 553: A library which wants to validate the md5 *and* doesn't support bep47/bep52 would need a valid MD5. Hashing all zeros seems fine?

src/Tests/Tests.MonoTorrent.Client/MonoTorrent/TorrentV2Test.cs
```diff
@@ -116,6 +116,7 @@ public void LoadV2OnlyTorrent ()
             Assert.AreEqual (InfoHash.FromHex ("caf1e1c30e81cb361b9ee167c4aa64228a7fa4fa9f6105232b28ad099f3a302e"), V2OnlyTorrent.InfoHashes.V2);
         }
 
+        // TODO: fails with MonoTorrent.TorrentException : the 'piece layers' dictionary did not contain an entry for the file 'tbl-tint.mpg'
```
alanmcgovern, line 119: I pushed a tiny change which fixes this. When zeroing out the padding for the last file in the torrent you had forgotten to copy over the `RootHash` property. This is what caused the file to not be found in the piece layers dictionary :)

alanmcgovern:
Overall this looks pretty good! The extra validations are great.

I've had a busy week or so in work and am flying out for a 2 week holiday on Saturday. As such I'm not likely to get back to you with reviews until I'm back :( 

I haven't had much time to thoroughly think through `PaddingAwareReadAsyncForCreator `, though from an initial review it feels like it could be the right approach. 

Thanks for the work so far, and i'm looking forward to getting this merged after i'm back!

src/MonoTorrent.Client/MonoTorrent/Torrent.cs
```diff
@@ -712,32 +760,45 @@ static PieceHashesV2 LoadHashesV2 (IList<ITorrentFile> files, BEncodedDictionary
...
+                    files.Add (new TorrentFile (path, length, 0, 0, 0, TorrentFileAttributes.None, 0));
                 } else {
                     totalPieces++;
-                    var offsetInTorrent = (files.LastOrDefault ()?.OffsetInTorrent ?? 0) + (files.LastOrDefault ()?.Length ?? 0);
+                    var offsetInTorrent = (files.LastOrDefault ()?.OffsetInTorrent ?? 0) + (files.LastOrDefault ()?.Length ?? 0) + (files.LastOrDefault ()?.Padding ?? 0);
                     var piecesRoot = data.TryGetValue ("pieces root", out var value) ? ((BEncodedString) value).AsMemory () : ReadOnlyMemory<byte>.Empty;
-                    files.Add (new TorrentFile (path, length, totalPieces, totalPieces + (int) (length / pieceLength), offsetInTorrent, piecesRoot));
+
+                    // TODO JMIK: wrong calculation for endIndex, when length is exactly pieceLength?
+                    files.Add (new TorrentFile (path,
```
alanmcgovern, line 775: Is it feasible to rewrite this method so it creates a bunch of `TorrentFileTuple` (just like the nice cleanup you did for the v1 method) and then both the v1 and v2 implementation could rely on the logic in `TorrentFile.Create` to set the right start/end piece indices etc?

src/MonoTorrent.Client/MonoTorrent/Torrent.cs
```diff
@@ -712,32 +760,45 @@ static PieceHashesV2 LoadHashesV2 (IList<ITorrentFile> files, BEncodedDictionary
...
         {
             var files = new List<ITorrentFile> ();
             int totalPieces = -1;
             foreach (var entry in fileTree)
-                LoadTorrentFilesV2 (entry.Key.Text, (BEncodedDictionary) entry.Value, files, pieceLength, ref totalPieces, "");
+                LoadTorrentFilesV2 (entry.Key.Text, (BEncodedDictionary) entry.Value, files, pieceLength, ref totalPieces, "", isHybrid);
+
+            // padding of last torrent must be 0.
+            var last = files.Last ();
+            files[files.Count - 1] = new TorrentFile (last.Path, last.Length, last.StartPieceIndex, last.EndPieceIndex, last.OffsetInTorrent, last.PiecesRoot, TorrentFileAttributes.None, 0);
```
alanmcgovern, line 801: I added `last.PiecesRoot` here to propagate that data!

jpmikkers:
Alan no worries on the review, have a great holiday!

jpmikkers:
I verified that md5 hashing is also broken on master, `file.MD5` is never set in `HashAllDataAsync()`.
See the test here: https://github.com/jpmikkers/monotorrent/commit/d2a8a8c297f9509bd4842acb5c9b69fd3e57cbd8

alanmcgovern:
@jpmikkers I fixed the MD5 issue last night - finally!

What's left to complete in this change? If it's pretty much done now I'll give it a final review and land any minor tweaks as needed before, or after, merging :)

jpmikkers:
I also had a small fix for the remaining // TODO: JMIK (the endindex rounding issue) but suddenly vs2020 doesn't let me push that commit, *groan*

alanmcgovern:
> totalPieces + (int) ((length - 1) / pieceLength) : totalPieces

https://github.com/alanmcgovern/monotorrent/pull/526/commits/96fa486754adaebbf6217df4a90ca2aafb433e24

You were right!

src/MonoTorrent.Client/MonoTorrent/TorrentCreator.cs
```diff
@@ -232,6 +234,11 @@ internal async Task<BEncodedDictionary> CreateAsync (string name, List<InputFile
             info["name"] = (BEncodedString) name;
             AddCommonStuff (torrent);
 
+            foreach (var file in files.Take (files.Count - 1)) {
+                file.Padding = (UsePadding && file.Length > 0) ? PieceLength - (file.Length % PieceLength) : 0;
```
alanmcgovern: I just realised that this would add `PieceLength ` padding if `file.Length` is an exact multiple of `PieceLength`.

I added a test and fixed the bug!

src/MonoTorrent.Client/MonoTorrent/TorrentFile.cs
```diff
@@ -109,40 +139,64 @@ internal static ITorrentFile[] Create (int pieceLength, params long[] lengths)
...
-                if ((totalSize + length) % pieceLength == 0)
-                    pieceEnd--;
-
-                if (length == 0) {
-                    pieceStart = i > 0 ? results[i - 1].StartPieceIndex : 0;
-                    pieceEnd = i > 0 ? results[i - 1].StartPieceIndex : 0;
-                    startOffsetInTorrent = i > 0 ? results[i - 1].OffsetInTorrent : 0;
+                // catch pathological case of too much padding
+                if (pieceEnd != pieceEndWithPadding) {
+                    throw new ArgumentException ("A file in the torrent has more padding than needed.");
```
alanmcgovern, line 184: This check was helpful as it triggered when I created the testcase to see what happened when `file.Length == PieceLength` :)

alanmcgovern:
I gave this a few rounds of checks and didn't find any more issues! Thanks for the great work! I'll let this continue baking as I (slowly) slipstream in the final bits for bep52 :D

Thanks for the patch!

PR #536, opened 2022-06-19
Use PiecePicker for arbitrary things

alanmcgovern:
PiecePicker has been simplified to have the concept of having a `Piece`, where each piece has 1 or more `Segments`.

The invariant is that once you download all the segments in 1 piece, the piece can be validated, allowing the data to be requested again if validation fails, or if it succeeds then the piece can be marked as such so it isn't requested again.

In concrete terms, there is no change to regular piece picking. However,

1) Fetching torrent metadata
The chunks are not independently verifiable so this data is downloaded as a single piece with N segments. Once all segments are received the data is hashed.

2) Fetching BitTorrent v2 hashes
The chunks are independently verifiable, so this data is represented as a series of pieces, each of which has 1 block. As soon as any block is received the data is verified.

This makes it much easier to double-request pieces, or to time out requests which aren't fulfilled due to a bad client improperly responding.


alanmcgovern:
TODO:
1) Make sure timeouts are respected
2) Make sure peer disconnects are respected

Issue #560, opened 2022-10-21
Empty file in torrent causes infinite loop


borigas:
Creating a v2 torrent from a folder with an empty file causes an infinite loop in `MerkleHash.TryHash()` because `src.Length == 0`. This test repros the problem in master and release-v3.0.0-beta-0034.
```
[TestFixture]
public class EmptyFileInTorrentTests
{
    [Test]
    public void EmptyFileInTorrent ()
    {
        var dir = new DirectoryInfo ("Data");
        if (!dir.Exists)
            dir.Create ();

        var emptyFile = new FileInfo (Path.Combine (dir.FullName, "Empty.file"));
        File.WriteAllText (emptyFile.FullName, "");

        var nonEmptyFile = new FileInfo (Path.Combine (dir.FullName, "NonEmpty.file"));
        File.WriteAllText (nonEmptyFile.FullName, "aoeuaoeu");

        emptyFile.Refresh ();
        Assert.AreEqual (0, emptyFile.Length);

        ITorrentFileSource fileSource = new TorrentFileSource (dir.FullName);
        TorrentCreator torrentCreator = new TorrentCreator (TorrentType.V1V2Hybrid);
        var creationTask = torrentCreator.CreateAsync (fileSource);

        bool completed = creationTask.Wait (10000);
        Assert.IsTrue (completed);

        var torrent = creationTask.Result;
        Assert.IsNotNull (torrent);
    }
}
```

I spent some time debugging, but I don't know enough about how empty files are supposed to be hashed to provide a fix. If there's more I can do to help diagnose, please let me know. Thanks for your help!

alanmcgovern:
Hah, whoops! I thought i had empty files covered, but obviously not for the V2 case!

In BitTorrent V1 empty files can exist anywhere in the torrent dictionary. MonoTorrent handles these by treating them as if they all existed at the start of the file (i.e. it sorts empty files so they're all at the start of `TorrentManager.Files` and `Torrent.Files` lists). This makes some of maths/calculations easier later on.

The same should be true for BitTorrent V2, with the additional support for completely skipping empty files when generating the V2  infohash. It should be more or less a 1 liner somewhere... let's see...

alanmcgovern:
This should be fixed in master! Let me know if you have any issues, and thanks for the bug report!

borigas:
Thanks for fixing the create with an empty file. That seems to be working as intended. I can create a torrent with an empty file now.

[...]

PR #565, opened 2022-10-26
More fixes for empty files

alanmcgovern:
Fixes https://github.com/alanmcgovern/monotorrent/issues/560

I'll split this up a bit and add a few more tests in other places!

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Modes/StartingMode.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Modes/StartingMode.cs
@@ -36,6 +36,8 @@ using System.Threading.Tasks;
 using MonoTorrent.BEncoding;
 using MonoTorrent.Logging;
 
+using ReusableTasks;
+
 namespace MonoTorrent.Client.Modes
 {
     class StartingMode : Mode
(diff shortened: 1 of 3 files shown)


src/MonoTorrent.Client/MonoTorrent.Client.Modes/StartingMode.cs
```diff
@@ -159,10 +159,16 @@ async Task VerifyHashState ()
             if (!Manager.HasMetadata || !Manager.HashChecked)
                 return;
 
-            // FIXME: I should really just ensure that zero length files always exist on disk. If the first file is
-            // a zero length file and someone deletes it after the first piece has been written to disk, it will
-            // never be recreated. If the downloaded data requires this file to exist, we have an issue.
+
             foreach (ITorrentManagerFile file in Manager.Files) {
+                if (file.Length == 0 && !File.Exists (file.FullPath)) {
+                    Directory.CreateDirectory (Path.GetDirectoryName (file.FullPath));
```
borigas, line 165: This is failing if the `TorrentManager` is started using `StartAsync()` instead of explicitly calling `CheckHashAsync()` because `Manager.HashChecked == false` and it short circuits above. The result is the empty file appears to only be created when explicitly checking the hash

src/MonoTorrent.Client/MonoTorrent.Client.Modes/StartingMode.cs
```diff
@@ -156,12 +177,9 @@ async Task TryLoadV2HashesFromCache()
         async Task VerifyHashState ()
         {
             // If we do not have metadata or the torrent needs a hash check, fast exit.
-            if (!Manager.HasMetadata || !Manager.HashChecked)
```
alanmcgovern, line 159: We can never reach this point *and* have `Manager.HasMetadata` return false:

https://github.com/alanmcgovern/monotorrent/pull/565/files#diff-ce0c0fad357facaa0ad04c3cdaec3735b709d7ec85caef1e6cf0f5e13183e6aaR61-R64

PR #566, opened 2022-10-27
Integration tests

borigas:
Here's the integration tests I previously copy/pasted on #560. I also added more to show the issue with implicit vs explicit hash check in #565. Let me know if I'm misunderstanding the intended behavior of `StartAsync()`

PR #567, opened 2022-11-03
Integration tests

borigas:
Here's the integration tests I previously copy/pasted on #560. I also added more to show the issue with implicit vs explicit hash check in #565.

After the changes from #565, these should all pass
image

This was originally in #566, but reopened to master instead of the deleted branch.

Thanks as always for a great library!


alanmcgovern (approved):
This looks good to me! These end-to-end tests are pretty helpful to catch misc issues which would otherwise be hard to think of when writing more focused unit tests. These tests already proved their worth by identifying several issues with how empty files were handled across the codebase :)

PR #576, opened 2022-11-16
Web Seed Integration Tests

borigas:
I think there may be a problem downloading via web seeds. Here's an integration test to illustrate the issue. I think the issue is downloading the last piece of the torrent. On the "bigger" 30KB files test, it downloads the full 30KB of the 1st file, but 16KB of 30KB in the 2nd.
image

Thanks for your help with this! And please let me know if PRing like this is how you'd like issues like this raised.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs
@@ -680,7 +680,7 @@ namespace MonoTorrent.Client.Modes
 
         void DownloadLogic (int counter)
         {
-            if (ClientEngine.SupportsWebSeed && (DateTime.Now - Manager.StartTime) > Settings.WebSeedDelay && Manager.Monitor.DownloadRate < Settings.WebSeedSpeedTrigger) {
+            if (ClientEngine.SupportsWebSeed && (DateTime.Now - Manager.StartTime) > Settings.WebSeedDelay && (Manager.Monitor.DownloadRate < Settings.WebSeedSpeedTrigger || Settings.WebSeedSpeedTrigger == 0)) {
                 foreach (Uri uri in Manager.Torrent!.HttpSeeds) {
                     var peer = new Peer (new PeerInfo (uri, CreatePeerId ()), Manager.InfoHashes.V1OrV2);
 
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs
@@ -29,6 +29,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Threading;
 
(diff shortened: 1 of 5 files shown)


borigas:
For me, the Microsoft.Net.Http nuget reference from fe13a59f is needed to build in net472

borigas:
Thanks for fixing those! New discovery: web seeds break on files larger than 2^17 bytes. I updated the bigger file test to show that

borigas:
It looks like if an HTTP request fails the http seed is completely removed and never retried. I added a test for this as well. I thought clearing the HttpSeeds at https://github.com/alanmcgovern/monotorrent/blob/master/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs#LL708-L710C52 may be to blame. I tried leaving the `HttpSeeds` intact and only adding a new peer if there were no peers, but the test still failed. I think I'm over my head on this one too.

borigas:
I think my idea on the retry issue was actually working. Just had some test teardown issues making it look like a failure. It's now checking if it needs to readd an HTTP seed based on the uri being in the list of peers and is passing the test. Hopefully there isn't other side effects from that change that I'm missing.

I still have no idea on the `2^17 + 1` issue though

alanmcgovern:
> For me, the Microsoft.Net.Http nuget reference from [fe13a59](https://github.com/alanmcgovern/monotorrent/commit/fe13a59fae50d713142fb21d139569e787f9488f) is needed to build in net472

Huh - that's odd! It builds for me without the reference. If it's needed please feel free to add it back in. It was causing some compile time warnings for newer framework versions so the reference might need something like: `Condition="$(TargetFramework) == "net472"` so it's ignored for newer framework versions then! I think there are places in the code where things are conditionally referenced like that already :) 

Good work fixing the retry issue, that was something which should've been fixed a while ago! I'm glad that bit of legacy/laziness has been fixed :) 

> I still have no idea on the 2^17 + 1 issue though

Hrm, i wonder if there's something happening where there's an off-by-one issue caused by files being an exact multiple of the block size... or something... I'll take a look!

borigas:
I should clarify on the 2^17+1 issue. Anything larger than 2^17 fails. I just trimmed the test down to the smallest size that would fail. I don't think it's an off by 1

alanmcgovern:
Fixed! The issue was that `DataStreamCount` wasn't being used to determine when we should stop reading from the data stream. Oops :)

alanmcgovern:
Interestingly, there are still some failures. I'll have to see if it's a race condition or a bug now :)

alanmcgovern:
Hah, embarrassing!

I didn't catch this before as I had lost the ability to run net472 tests until I had updated to https://github.com/alanmcgovern/monotorrent/pull/576/commits/9998ce87b908bc5dbb93c43b615e04616fea8d93 *and* remembered to add them back into the test playlist I was using! :)

It looks like things are still failing in CI for non-test related reasons:
```
The test source file "D:\a\1\s\src\Tests\Tests.MonoTorrent.IntegrationTests\bin\Debug\net6.0\Tests.MonoTorrent.IntegrationTests.dll" provided was not found.
##[error]Error: The process 'C:\Program Files\dotnet\dotnet.exe' failed with exit code 1
```

Maybe something has borked in the pipeline and it's not building the test projects now, or maybe it's building 'release' versions but trying to run 'debug' files (which weren't compiled). That's a problem for later :)

alanmcgovern:
@borigas Does this branch seem good to you? Does it solve the problems you were having with web seeds?

borigas:
I was having still having trouble after the overread fix yesterday, but I haven't had a chance to isolate it yet. It could be something wrong in my test. I'm ok merging this and opening a new one if needed, or letting this sit until I can investigate more and get back to you. Either way works. Thanks for all your help!

alanmcgovern:
SGTM! I'll merge this now and we can iterate some more if there are further issues :)

Thanks for the geaat tests and fixes!

PR #577, opened 2022-11-18
Start web seed if the speed trigger is 0

borigas:
From EngineSettingsBuilder.cs
```
        /// <summary>
        /// The download speed under which a torrent will start using web seeds. A value of
        /// 0 means webseeds will be added regardless of how fast the torrent is downloading.
        /// </summary>
        public int WebSeedSpeedTrigger {
```


alanmcgovern:
Whoops! Good catch!

PR #578, opened 2022-11-25
[tests] Close the response after writing the status code

alanmcgovern:
If the response is not closed then the connection will be treated as failed only after the 30 second timeout is hit.

The `WebSeedDownload_V1V2_RetryWebSeeder` test took 32 seconds to execute before this change. Now it takes 2 seconds.

### DIFF
--- a/src/Tests/Tests.MonoTorrent.IntegrationTests/IntegrationTests.cs
+++ b/src/Tests/Tests.MonoTorrent.IntegrationTests/IntegrationTests.cs
@@ -20,7 +20,7 @@ namespace Tests.MonoTorrent.IntegrationTests
     [TestFixture]
     public class IntegrationTests
     {
-        static TimeSpan CancellationTimeout = Debugger.IsAttached ? Timeout.InfiniteTimeSpan : TimeSpan.FromSeconds (10);
+        static readonly TimeSpan CancellationTimeout = Debugger.IsAttached ? Timeout.InfiniteTimeSpan : TimeSpan.FromSeconds (20);
 
         [OneTimeSetUp]
         public void FixtureSetup ()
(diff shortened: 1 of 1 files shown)


src/Tests/Tests.MonoTorrent.IntegrationTests/IntegrationTests.cs
```diff
@@ -250,11 +247,18 @@ private void OnHttpContext (IAsyncResult ar)
             if (_failHttpRequest) {
                 _failHttpRequest = false;
                 ctx.Response.StatusCode = 500;
-            } else if (localPath.Contains (relativeSeedingPath)) {
+                ctx.Response.Close ();
+            } else if (!localPath.Contains (relativeSeedingPath)) {
+                ctx.Response.StatusCode = 404;
+                ctx.Response.Close ();
```
alanmcgovern, line 253: @borigas The missing `Response.Close` call is why the `Retry` test needed a much longer execute.

src/Tests/Tests.MonoTorrent.IntegrationTests/IntegrationTests.cs
```diff
@@ -266,15 +270,16 @@ private void OnHttpContext (IAsyncResult ar)
...
-                    ctx.Response.OutputStream.Write (buffer, 0, buffer.Length);
-                    ctx.Response.OutputStream.Close ();
-                    return;
+                    if (fs.Read (buffer, 0, buffer.Length) == buffer.Length) {
+                        ctx.Response.OutputStream.Write (buffer, 0, buffer.Length);
+                        ctx.Response.OutputStream.Close ();
+                    } else {
+                        ctx.Response.StatusCode = 405;
+                        ctx.Response.Close ();
+                    }
```
alanmcgovern, line 279: I added a few extra checks here in case fs.Read doesn't read the right number of bytes, or `file == null` etc. I don't think the tests actually hit these issues anyway :P

PR #583, opened 2022-12-18
[core] Allow creation dates with a length of 0

winterqt:
Some torrents contain a creation date key with a zero-length string as its value. Currently, MonoTorrent will throw an exception when trying to parse these.

Let me know if there's another way you'd want to handle this (I'm guessing invalid) case. I'm not sure if a tool creates these or what, but I've seen it enough that I think something should be done here. I know torrent clients don't choke on this like MonoTorrent does, so maybe we can handle this more generically (e.g. skip parsing a key if its value's length is 0) rather than special casing the creation date.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent/Torrent.cs
+++ b/src/MonoTorrent.Client/MonoTorrent/Torrent.cs
@@ -531,6 +531,9 @@ namespace MonoTorrent
                         break;
 
                     case ("creation date"):
+                        if (keypair.Value.ToString () == String.Empty)
+                            break;
+
                         try {
                             try {
                                 CreationDate = UnixEpoch.AddSeconds (long.Parse (keypair.Value.ToString ()!));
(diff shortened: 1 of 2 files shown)


alanmcgovern:
I'm +1 on this a fix for the specific issue of creation dates as it cannot cause side-effects. 

Whether or not this can be added as a generic fix probably depends on whether or not anything in the Torrent parsing code checks for the existence of specific keys without also ensuring the value is valid. 

If there are no occurrences of this, I'd be +1 on making this a generic fix, where all empty keys are quietly skipped. I haven't reviewed this yet.  

I did do a quick review to see if this could cause a user visible change, but didn't see anywhere that would occur :)

alanmcgovern:
Would you be up for augmenting this to skip all keys whose values are bencodedstrings and those strings are empty?

(I think that's the only case that matters?)

winterqt:
Sure, closing in favor of #584 due to more fitting branch name.

PR #614, opened 2023-01-30
Integration tests ipv6

alanmcgovern:
Finalize ipv6 support by adding end-to-end tests using ipv6.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Tracker/TrackerRequestFactory.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Tracker/TrackerRequestFactory.cs
@@ -56,8 +56,8 @@ namespace MonoTorrent.Trackers
             Func<string, (string?, int)> reportedAddressFunc = type => {
                 // IPV6 support - If we need to do an ipv4 announce *and* an ipv6 announce, where's the best place for this to live?
                 string? ip = null;
-                int port = engine.GetOverrideOrActualListenPort ("ipv6").GetValueOrDefault (-1);
-                if (engine.Settings.ReportedListenEndPoints.TryGetValue ("ipv6", out var reportedAddress))
+                int port = engine.GetOverrideOrActualListenPort (type).GetValueOrDefault (-1);
+                if (engine.Settings.ReportedListenEndPoints.TryGetValue (type, out var reportedAddress))
                     ip = reportedAddress.Address.ToString ();
                 return (ip, port);
             };
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Tracker/TrackerRequestFactory.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Tracker/TrackerRequestFactory.cs
@@ -27,6 +27,8 @@
 //
 
 
+using System;
+
 using MonoTorrent.Client;
 using MonoTorrent.Connections;
 
@@ -51,11 +53,14 @@ namespace MonoTorrent.Trackers
             requireEncryption = requireEncryption && ClientEngine.SupportsEncryption;
             supportsEncryption = supportsEncryption && ClientEngine.SupportsEncryption;
 
-            // IPV6 support - If we need to do an ipv4 announce *and* an ipv6 announce, where's the best place for this to live?
-            string? ip = null;
-            int port = engine.GetOverrideOrActualListenPort ("ipv4").GetValueOrDefault (-1);
-            if (engine.Settings.ReportedListenEndPoints.TryGetValue ("ipv4", out var reportedAddress))
-                ip = reportedAddress.Address.ToString ();
+            Func<string, (string?, int)> reportedAddressFunc = type => {
+                // IPV6 support - If we need to do an ipv4 announce *and* an ipv6 announce, where's the best place for this to live?
+                string? ip = null;
+                int port = engine.GetOverrideOrActualListenPort ("ipv6").GetValueOrDefault (-1);
+                if (engine.Settings.ReportedListenEndPoints.TryGetValue ("ipv6", out var reportedAddress))
+                    ip = reportedAddress.Address.ToString ();
+                return (ip, port);
+            };
 
             // FIXME: In metadata mode we need to pretend we need to download data otherwise
             // tracker optimisations might result in no peers being sent back.
@@ -67,7 +72,7 @@ namespace MonoTorrent.Trackers
                                           Manager.Monitor.DataBytesSent,
                                           bytesLeft,
                                           clientEvent, Manager.InfoHashes, requireEncryption, Manager.Engine!.PeerId.AsMemory (),
-                                          ip, port, supportsEncryption);
+                                          reportedAddressFunc, supportsEncryption);
         }
 
         public ScrapeRequest CreateScrape ()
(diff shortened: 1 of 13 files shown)


ycy1164656:
actually, I tested every time you push a commit /lol

This time I saw you fixed 

`var listener = new Socket (OriginalEndPoint.AddressFamily, SocketType.Stream, ProtocolType.Tcp);`

in ../src/MonoTorrent.Connections/MonoTorrent.Connections.Peer/PeerConnectionListener.cs

and 

` uri = socket.AddressFamily switch {
                    AddressFamily.InterNetwork => new Uri ($"ipv4://{endpoint}"),
                    AddressFamily.InterNetworkV6 => new Uri ($"ipv6://{endpoint}"),
                    _ => throw new NotSupportedException ($"AddressFamily.{socket.AddressFamily} is unsupported")
                };`

in ../src/MonoTorrent.Connections/MonoTorrent.Connections.Peer/SocketPeerConnection.cs

and 

` Func<string, (string?, int)> reportedAddressFunc = type => {
                // IPV6 support - If we need to do an ipv4 announce *and* an ipv6 announce, where's the best place for this to live?
                string? ip = null;
                int port = engine.GetOverrideOrActualListenPort (type).GetValueOrDefault (-1);
                if (engine.Settings.ReportedListenEndPoints.TryGetValue (type, out var reportedAddress))
                    ip = reportedAddress.Address.ToString ();
                return (ip, port);
            };`
in ../src/MonoTorrent.Client/MonoTorrent.Client.Tracker/TrackerRequestFactory.cs; 

I run clientSample on two mac,but still can not show ipv6 peers, so  I edit 

`var connectionUri = new Uri ($"ipv4://{dict[IPKey]}:{dict[PortKey]}");`

to 

`Uri connectionUri;
            if ($"{dict[IPKey]}".Length > 15)
            {
                connectionUri = new Uri ($"ipv6://[{dict[IPKey]}]:{dict[PortKey]}");
            }
            else
            {
               connectionUri = new Uri ($"ipv4://{dict[IPKey]}:{dict[PortKey]}");
            }`

in ../src/MonoTorrent.Client/MonoTorrent.Client/Peers/PeerDecoder.cs

,now I can get those peers list,but still can not connect to each other;

as you can see,

image

port 12346 is mac 1 which is local device,
port 22345 is mac 2 which is remote device,
both device are running monotorrent clientsample,
and port 56185 come from local device but I don't know where it generated.

then,connection failed

image

and if I add 
`AllowedEncryption = new List<EncryptionType> { EncryptionType.PlainText },`

the result would be 

image


and on the remote device ,I saw these weird ports which I didn't set to engine

image

.

Again, really appreciate your work!

alanmcgovern:
Bonkers! I'll give that another shot and see if I can replicate your issue locally :s

alanmcgovern:
huh

I took two different physical machines, used a build created from https://github.com/alanmcgovern/monotorrent/pull/615/files (you can see how I hardcoded IP addresses for my setup, and also explicitly disabled ipv4 from monotorrent's perspective) and was able to transfer data from one machine to the other.

To debug your issue further i'll probably need logs from your software. Could you 


Also..
> Uri connectionUri; if ($"{dict[IPKey]}".Length > 15) { connectionUri = new Uri ($"ipv6://[{dict[IPKey]}]:{dict[PortKey]}"); } else { connectionUri = new Uri ($"ipv4://{dict[IPKey]}:{dict[PortKey]}"); }

This looks like something i may need to fix! Let me look into that too :)

alanmcgovern:
https://github.com/alanmcgovern/monotorrent/pull/616

alanmcgovern:
> and on the remote device ,I saw these weird ports which I didn't set to engine

Outgoing connections aren't bound to a specific port, so it's "normal" for the peer receiving the connection to see the 'right' IP address for the remote peer, but a random 'port'.

ycy1164656:
I tested the master branch code ,only changed port on both devices,here is the torrent file

[5C268BB3F8DC15A32A685B116917E4178F0D7C70.txt](https://github.com/alanmcgovern/monotorrent/files/10552963/5C268BB3F8DC15A32A685B116917E4178F0D7C70.txt)

change the .txt to .torrent you will get it;

And these are logs on both device

[20230201_log.txt](https://github.com/alanmcgovern/monotorrent/files/10553011/20230201_log.txt)


[20230201_log.txt](https://github.com/alanmcgovern/monotorrent/files/10553013/20230201_log.txt)


If you have a proper ipv6 environment, you can use that torrent,I'll keep my pc open and seeding til 17:30 Beijing .

alanmcgovern:
```
2/1/2023 10:25:30 AM      INFO: Mode:ipv6://[240e:476:303:6806:7dbe:1a12:c4e1:eda5]:54321/: HandShake.Handle - Invalid peerid
```

this is the problem! Phew! I'm glad the logging caught it.

The line in question is:
https://github.com/alanmcgovern/monotorrent/blob/99b5f9f24fdb7383c7da3c9d3ff7791dd21f8eef/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs#L224-L230

And the issue is that the peer id reported by the tracker is not the same as the peer id the client sends when it establishes the connection. This logic is only in effect for private torrents. This particular log line should be augmented to include the expected, and actual, peerids to understand what's going on.

I'll give a shout when i have something to test, but in the meantime you don't need to continue seeding. it could be a day or two before i've time to focus on this!

ycy1164656:
`Are you running multiple clients behind the same public IP and the tracker isn't disambiguating those?`
Yes, I am running multiple clients, but,as for ipv6, I think there is no concept like "behind public Ip",lol

`Are you restarting clients frequently? `

Yes I restarted clients often for testing.

`Is this just a bug in monotorrent maybe?`

Same torrent same tracker works well when using qbittorrent or bitcomet, so...it may be a monotorrent exclusive bug ...lol

alanmcgovern:
Huh, well, i'm not going to be able to independently verify what your tracker is returning :( 

What i can do is augment logging to better capture this scenario, and just remove the explicit error when there's a `peer_id` mismatch.

alanmcgovern:
@ycy1164656 I just released 3.0.0-beta.rev0157 which contains the logging improvement mentioned above. I'd love if you could share the logs again (especially if you restart clients in between) to understand exactly what's happening using this version :)

edit: I then released 3.0.0-beta-rev0159 with a little bit more logging. If you can, use that version ;)

ycy1164656:
> @ycy1164656 I just released 3.0.0-beta.rev0157 which contains the logging improvement mentioned above. I'd love if you could share the logs again (especially if you restart clients in between) to understand exactly what's happening using this version :)
> 
> edit: I then released 3.0.0-beta-rev0159 with a little bit more logging. If you can, use that version ;)

alright, I'll test it next week :)

alanmcgovern:
I'm disabling this particular check by default ( https://github.com/alanmcgovern/monotorrent/pull/628 ) . While it can still be enabled as-needed, i can't really think of a situation where peer id checks (as described by the spec) are useful nowadays, especially since other clients randomise these IDs anyway :)

ycy1164656:
> I'm disabling this particular check by default ( #628 ) . While it can still be enabled as-needed, i can't really think of a situation where peer id checks (as described by the spec) are useful nowadays, especially since other clients randomise these IDs anyway :)

I tested the master branch code hour ago,and it's done a perfect seeding job!
Great !

PR #617, opened 2023-01-31
[core] Ratelimiters should allow requests larger than chunk size

alanmcgovern:
If the ratelimiter is set to download at 15kB/sec or less, the amount of data in the ratelimit will always be less than the size of one block. In this case, we should allow the request to succeed.

Once there's negative data left in the allowance, deny requests. As the ratelimiter is topped up over time, future requests will be unblocked.

### DIFF
--- a/src/Tests/Tests.MonoTorrent.Client/MonoTorrent.Client.RateLimiters/RateLimiterTests.cs
+++ b/src/Tests/Tests.MonoTorrent.Client/MonoTorrent.Client.RateLimiters/RateLimiterTests.cs
@@ -1,0 +1,64 @@
+﻿//
+// RateLimiterTests.cs
+//
+// Authors:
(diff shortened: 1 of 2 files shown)


src/Tests/Tests.MonoTorrent.Client/MonoTorrent.Client.RateLimiters/RateLimiterTests.cs
```diff
@@ -0,0 +1,64 @@
+﻿//
+// RateLimiterTests.cs
+//
+// Authors:
+//   Alan McGovern alan.mcgovern@gmail.com
+//
+// Copyright (C) 2019 Alan McGovern
```
alanmcgovern, line 7:
```suggestion
// Copyright (C) 2023 Alan McGovern
```

PR #645, opened 2023-07-07
Backported use of UrlEncode/UrlDecode from HttpUtility in Infohash

mr-const:
UriHelper uses outdated standard and makes hashes which sometimes are not accepted by trackers, for example if infohash contains 0x20 (space) it's substituted with '+' sign.

### DIFF
--- a/src/MonoTorrent/MonoTorrent/InfoHash.cs
+++ b/src/MonoTorrent/MonoTorrent/InfoHash.cs
@@ -31,6 +31,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
+using System.Web;
 
 namespace MonoTorrent
 {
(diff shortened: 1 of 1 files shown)


src/MonoTorrent/MonoTorrent/InfoHash.cs
```diff
@@ -98,7 +99,7 @@ public string ToHex ()
 
         public string UrlEncode ()
         {
-            return UriHelper.UrlEncode (Hash);
+            return HttpUtility.UrlEncode (Hash).Replace ("+", "%20");
```
alanmcgovern, line 102: This looks like a reasonable change if it improves compatibility, through `+` in the query part is supposed to be supported :s

alanmcgovern:
Hrmm, i assume this holds true for all other places where this urlencodes - magnet links, tracker ids, etc. i'lll do a quick pass on this to find more places

PR #664, opened 2024-07-22
Embed debug info into DLLs

lostmsu:
This greatly simplifies debugging.

These days storage is cheap, and debug symbols are small.

Removes the need to upload debug symbols separately, setting up debuggers to download them from NuGet (or you name it), etc etc etc


alanmcgovern (approved):
SGTM. 

I could've sworn I did this already, but maybe it was for another project. 

Thanks!

PR #665, opened 2024-07-31
Fix two issues affecting webseeds

alanmcgovern:
1) Ensure web requests request data in-order by fixing the piecepicker
   to correctly use 'CanRequestMorePieces'

2) Fix a bug in the handling of padding files. A request for some data
   from a webseed needs to take padding files into account in all cases.
   A request can begin or end inside padding, it can also begin/end in
   regular files while optionally crossing one stretch of padding. This
   means the existing logic to insert synethic requests to the padding
   file are broken in a number of cases.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.Connections/HttpPeerConnection.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Connections/HttpPeerConnection.cs
@@ -337,45 +337,51 @@ namespace MonoTorrent.Connections.Peer
 
         void CreateWebRequests (BlockInfo start, BlockInfo end)
         {
-            // Properly handle the case where we have multiple files
-            // This is only implemented for single file torrents
             Uri uri = Uri;
 
             if (Uri.OriginalString.EndsWith ("/"))
                 uri = new Uri (uri, $"{TorrentData.Name}/");
 
             // startOffset and endOffset are *inclusive*. I need to subtract '1' from the end index so that i
-            // stop at the correct byte when requesting the byte ranges from the server
+            // stop at the correct byte when requesting the byte ranges from the server.
+            //
+            // These values are also always relative to the *current* file as we iterate through the list of files.
             long startOffset = TorrentData.TorrentInfo!.PieceIndexToByteOffset (start.PieceIndex) + start.StartOffset;
-            long endOffset = TorrentData.TorrentInfo!.PieceIndexToByteOffset (end.PieceIndex) + end.StartOffset + end.RequestLength;
+            long count = TorrentData.TorrentInfo!.PieceIndexToByteOffset (end.PieceIndex) + end.StartOffset + end.RequestLength - startOffset;
 
             foreach (var file in TorrentData.Files) {
+                // Bail out after we've read all the data.
+                if (count == 0)
+                    break;
+
+                var lengthWithPadding = file.Length + file.Padding;
+                // If the first byte of data is from the next file, move to the next file immediately
+                // and adjust start offset to be relative to that file.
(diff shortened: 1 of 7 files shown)


src/MonoTorrent.Client/MonoTorrent.Client/PeerId.PieceRequester.cs
```diff
@@ -61,6 +61,7 @@ public partial class PeerId : IRequester
 
         int IRequester.PreferredRequestAmount (int pieceLength)
         {
+            return 1;
```
alanmcgovern, line 64: This made it easier to trigger the bug. 8 pieces were being requested instead of 1, which was weird.

src/MonoTorrent.PiecePicking/MonoTorrent.PiecePicking/StandardPieceRequester.cs
```diff
@@ -137,7 +135,7 @@ public void AddRequests (IRequester peer, ReadOnlyBitField available, ReadOnlySp
             if (!peer.IsChoking && peer.AmRequestingPiecesCount == 0) {
                 ReadOnlyBitField? filtered = null;
                 PieceSegment segment;
-                while (peer.AmRequestingPiecesCount < maxRequests) {
+                while (peer.CanRequestMorePieces) {
```
alanmcgovern, line 138: This was what caused files to be picked out of order - effectively it always tried to add `MaxPendingRequests`, rather than relying on 'CanRequestMore'. Urgh

src/MonoTorrent.Client/MonoTorrent.Connections/HttpPeerConnection.cs
```diff
@@ -357,23 +357,41 @@ void CreateWebRequests (BlockInfo start, BlockInfo end)
...
                 }
                 // We want data from the end of the current file and from the next few files
-                else if (endOffset >= file.Length) {
-                    WebRequests.Enqueue ((u, startOffset, file.Length - startOffset));
-                    startOffset = 0;
-                    endOffset -= file.Length;
+                else if (endOffset >= (file.Length + file.Padding)) {
+                    // Maybe we need data from the actaul file?
+                    if (file.Length - startOffset < 0) {
+                        WebRequests.Enqueue ((u, startOffset, file.Length - startOffset));
```
borigas, line 368: If `file.Length - startOffset < 0`, I think we end up requesting a negative number of bytes here. Is that right?

borigas (commented):
Thanks for fixing this. I appreciate that you want to fix the root causes.

I don't know the details of the library enough to offer a great review, but I did notice a possible bug.

src/MonoTorrent.Client/MonoTorrent.Connections/HttpPeerConnection.cs
alanmcgovern: Ah, I had added a bunch of "throw" calls for all the negative ranges so I could catch the issue at instantiation time, not later on when the web request was used.

This branch is very much a WIP and doesn't pass tests yet - though that's still a good catch! There's a good chance I wouldn't notice that immediately and would waste a bunch of time trying to get the tests green 😆

src/Tests/Tests.MonoTorrent.PiecePicking/StandardPickerTests.cs
```diff
@@ -68,7 +68,7 @@ public void Setup ()
                 peers[i] = new BitField (pieceCount);
             }
         }
-
+
```
alanmcgovern, line 71:
```suggestion

```

Issue #672, opened 2024-08-09
From `FetchingHashes` to `Downloading` for V2 is so slow


peerlessDJ:
```
TorrentManager torrentManager = await _clientEngine.AddAsync(pArgs.Torrent, pArgs.SavePath);
foreach (var file in torrentManager.Files)
{
    var torrentItem = pArgs.TorrentItems.FirstOrDefault(item => item.TorrentFile.Path == file.Path);
    await torrentManager.SetFilePriorityAsync(file, torrentItem != null ? (Priority)torrentItem.Priority : Priority.DoNotDownload);
}

torrentManager.TorrentStateChanged += delegate (object o, TorrentStateChangedEventArgs e) 
{
    Debug.WriteLine($"TorrentStateChanged----OldState: [{e.OldState}]   NewState: [{e.NewState}]  DateTime：[{DateTime.Now.ToString()}]");
};

await torrentManager.StartAsync();
```

The log is 
```
TorrentStateChanged----OldState: [Stopped]   NewState: [Starting]  DateTime：[2024/8/9 15:03:18]
TorrentStateChanged----OldState: [Starting]   NewState: [Hashing]  DateTime：[2024/8/9 15:03:18]
TorrentStateChanged----OldState: [Hashing]   NewState: [FetchingHashes]  DateTime：[2024/8/9 15:03:18]
TorrentStateChanged----OldState: [FetchingHashes]   NewState: [Downloading]  DateTime：[2024/8/9 15:05:28]
```
from FetchingHashes to Downloading wait 2 minutes

alanmcgovern:
Ah - I think I see the issue. 

There's a bug whereby hash request messages are enqueued to send to a peer, but the never begins processing, and so the requests are left there until the timeout is reached.
 
![whoops](https://github.com/user-attachments/assets/57085986-c38b-4e57-89f8-8265de68927c)

PR #673, opened 2024-08-10
Ensure pending messages are processed from the send queue regularly

alanmcgovern:
This fixes a race condition where peers would sometimes get stuck while fetching bittorrent v2 hashes, depending on the timing of when the first request was made and then the final part of the handshake completed.

If requests were enqueued early, everything was fine. If requests were queued late, then nothing would kick off the queue consumer and enqueued messages would be left pending until the peer timed out, or until a KeepAlive message was sent which would unblock the queue.

Addresses issues like https://github.com/alanmcgovern/monotorrent/issues/672

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Modes/Mode.cs
@@ -613,6 +613,9 @@ namespace MonoTorrent.Client.Modes
 
         protected void PreLogicTick (int counter)
         {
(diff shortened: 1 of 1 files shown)


peerlessDJ:
emmm，is there no way to optimize it further? The first time to download from torrent file, it always get this problem

alanmcgovern:
That depends on what you mean. If you're wondering if hashes can be fetched faster - try the fix first.

If there's another issue - could you open a new issue describing the particular performance issue you experience?

PR #674, opened 2024-08-10
Ensure hash checking a file is a purely read-only operation

alanmcgovern:
The original implementation had some conveniences which are not great in certain circumstances. The DiskManager implementation would do it's utmost to ensure files were always the correct size/length, and that zero length files always existed on disk.

This is fine *except* if someone wishes to check if a particular file matches a particular hash - in this scenario, while they simply wanted to see if there was a match, the act of running the hashcheck would also truncate the file if it were too large.

Now the engine doesn't do this, and `TorrentManager.HashCheckAsync` neither truncates, nor creates zero length files.

If someone takes the step of actually calling `TorrentManager.StartAsync`, this is a clear sign they *really do* want to download the torrent, and at that point zero length files will be created, and existing files will be truncated if they are too large. This operation occurs only during the 'Starting' phase, prior to enter 'Downloading' or 'Seeding'. This makes things a little easier to reason about :) 

However, to support this in the optimal way additional methods need to be added to some of the interfaces used for IO, such as a get/set length method.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Modes/DownloadMode.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Modes/DownloadMode.cs
@@ -88,8 +88,6 @@ namespace MonoTorrent.Client.Modes
 
         internal async Task UpdateSeedingDownloadingState ()
         {
-            // FIXME: Ensure this properly handles the case where files are missing.
-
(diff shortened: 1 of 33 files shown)


ManlyMarco:
This is great, it should fix all of my issues. The only thing that would make it better is letting HashCheckAsync use fast resume data.

PR #679, opened 2024-08-20
Expose IPeerConnectionListeners publicly

alanmcgovern:
This makes it easier for everyone to determine which specific local ports are bound when '0' is used as the preferred port.


src/MonoTorrent.Client/MonoTorrent.Client/ClientEngine.cs
```diff
@@ -231,7 +231,12 @@ public async Task SaveStateAsync (string pathToStateFile)
 
         internal Factories Factories { get; }
 
-        internal IList<IPeerConnectionListener> PeerListeners { get; set; } = Array.Empty<IPeerConnectionListener> ();
+
+        /// <summary>
+        /// A readonly list of the listeners which the engine is using to receive incoming connections from other peers.
+        /// This are created by passing <see cref="EngineSettings.ListenEndPoints"/> to the <see cref="Factories.CreatePeerConnectionListener(IPEndPoint)"/> factory method.
+        /// </summary>
+        public IList<IPeerConnectionListener> PeerListeners { get; set; } = Array.Empty<IPeerConnectionListener> ();
```
alanmcgovern, line 239:
```suggestion
        public IList<IPeerConnectionListener> PeerListeners { get; private set; } = Array.Empty<IPeerConnectionListener> ();
```

PR #691, opened 2024-09-02
Check and ban peer when accepting incoming connection

kaedei:
### Description

The current implementation only checks if a peer should be banned during `TryConnect`. This PR adds a check using `ShouldBanPeer()` within the `IncomingConnectionAcceptedAsync()` method to ensure that peers are also evaluated for banning when an incoming connection is accepted.

### Changes

- Added `ShouldBanPeer()` check in `IncomingConnectionAcceptedAsync()`.

### Motivation

Enhancing the robustness of the peer banning mechanism by ensuring that peers are evaluated for banning not only during connection attempts but also when incoming connections are accepted.


kaedei:
Sorry, selected wrong branch. I will create a new PR for this.

PR #716, opened 2025-02-01
exception in SimpleTorrentManager.GetPeers

### LINKED ISSUE #715: Exception in SimpleTorrentManager.GetPeers
I have a tracker server that handles many announces in quick succession.

```
System.ArgumentException: Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection.
   at System.Buffer.BlockCopy(Array src, Int32 srcOffset, Array dst, Int32 dstOffset, Int32 count)
   at MonoTorrent.TrackerServer.SimpleTorrentManager.GetPeers(BEncodedDictionary response, Int32 count, Boolean compact, AddressFamily addressFamily) in C:\Projects\monotorrent\src\MonoTorrent.Client\MonoTorrent.TrackerServer\SimpleTorrentManager.cs:line 179
   at MonoTorrent.TrackerServer.TrackerServer.ListenerReceivedAnnounce(Object sender, AnnounceRequest e) in C:\Projects\monotorrent\src\MonoTorrent.Client\MonoTorrent.TrackerServer\TrackerServer.cs:line 268
   at MonoTorrent.Connections.TrackerServer.TrackerListener.Handle(NameValueCollection collection, IPAddress remoteAddress, Boolean isScrape) in C:\Projects\monotorrent\src\MonoTorrent.Trackers\MonoTorrent.Connections.TrackerServer\TrackerListener.cs:line 65
```

It appears to happen if you announce an ipv6 address then immediately announce the same torrent with an ipv4 address. The PeerList is remains the same length thus is not updated but the address family switches so the ipv6 peerlist will overflow the the buffer which was allocated for ipv4 stride sizes.

https://github.com/alanmcgovern/monotorrent/blob/9e98a44c3af93ace7fe11da363fe345a60c0c93f/src/MonoTorrent.Client/MonoTorrent.TrackerServer/SimpleTorrentManager.cs#L155

https://github.com/alanmcgovern/monotorrent/blob/9e98a44c3af93ace7fe11da363fe345a60c0c93f/src/MonoTorrent.Client/MonoTorrent.TrackerServer/SimpleTorrentManager.cs#L172

https://github.com/alanmcgovern/monotorrent/blob/9e98a44c3af93ace7fe11da363fe345a60c0c93f/src/MonoTorrent.Client/MonoTorrent.TrackerServer/SimpleTorrentManager.cs#L179

johnsu:
Fixes #715: Buffer overflow when announcing IPv4 address immediately after IPv6

This PR addresses an issue where announcing an IPv4 address immediately after an IPv6 address for the same torrent could cause an unhandled exception. The problem occurs because:

The `PeersList` is not updated when its length remains the same, even if the address family changes.
This leads to a mismatch between the address family of the peers in `PeersList` and the newly announced address family.
Consequently, when trying to copy IPv6 peer data (18 bytes) into a buffer allocated for IPv4 peers (6 bytes), a System.ArgumentException occurs during Buffer.BlockCopy.

The fix involves updating the condition for refreshing the `PeersList`. Now, it checks both the count and the address family of the peers, ensuring that the list is always consistent with the current announcement type.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.TrackerServer/SimpleTorrentManager.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.TrackerServer/SimpleTorrentManager.cs
@@ -169,8 +169,12 @@ namespace MonoTorrent.TrackerServer
             int start = Random.Next (0, peers.Count);
 
             lock (PeersList) {
-                if (PeersList.Count != peers.Values.Count)
-                    PeersList = new List<Peer> (peers.Values);
+                // Update PeersList only if the count differs or if the address family has changed
+                if (PeersList.Count != peers.Values.Count || 
+                    (PeersList.Count > 0 && PeersList[0].ClientAddress.AddressFamily != addressFamily))
+                {
+                    PeersList = new List<Peer>(peers.Values);
+                }
             }
 
             while (total > 0) {


jeffward01:
@alanmcgovern bump 😄

alanmcgovern (approved):
SGTM, thanks!

PR #719, opened 2025-08-30
Fix: `MagnetLink.ConvertToString` should include Exact Length if known

IsaMorphic:
This PR addresses a key issue in the string conversion code for the `MagnetLink` class where the Exact Length `xl=` attribute would not be printed, even if the current instance provided such a value. The new code makes sure to print this parameter alongside the rest.


alanmcgovern (approved):
Thanks!

PR #720, opened 2025-08-30
Fixes for Android & Other OSes using Unix-like file paths

### LINKED ISSUE #708: DERP
Getting `InvalidOperationExcepttion` on Linux (NixOS) with master branch:
```csharp
var source = new TorrentFileSource("/tmp/somedir");
await new TorrentCreator(TorrentType.V2Only).CreateAsync(source, CancellationToken.None);
```

`InvalidOperationException ("DERP")`

image

`file.Source == "/tmp/somedir/README.txt"`

`Path.AltDirectorySeparatorChar == '/'`

IsaMorphic:
This PR fixes https://github.com/alanmcgovern/monotorrent/issues/708 for OS platforms that use a forward-slash as their path separator character. Tested and working on Android and iOS platforms, Linux should be fine too.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent/TorrentCreator.cs
+++ b/src/MonoTorrent.Client/MonoTorrent/TorrentCreator.cs
@@ -62,7 +62,7 @@ namespace MonoTorrent
             public IList<ITorrentFile> Files { get; }
             public InfoHashes InfoHashes { get; }
             public string Name => "";
-            public int PieceLength { get;}
+            public int PieceLength { get; }
             public long Size { get; }
 
             public TorrentInfo (InfoHashes infoHashes, IList<ITorrentFile> files, int pieceLength)
@@ -166,7 +166,7 @@ namespace MonoTorrent
         public TorrentCreator (TorrentType type)
             : this (type, Factories.Default)
         {
-            
+
         }
         public TorrentCreator (Factories factories)
             : this (TorrentType.V1V2Hybrid, factories)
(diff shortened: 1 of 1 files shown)


src/MonoTorrent.Client/MonoTorrent/TorrentCreator.cs
```diff
@@ -249,9 +249,11 @@ internal Task<BEncodedDictionary> CreateAsync (string name, ITorrentFileSource f
         internal async Task<BEncodedDictionary> CreateAsync (string name, ITorrentFileSource fileSource, CancellationToken token)
         {
             var source = fileSource.Files.ToList ();
-            foreach (var file in source)
-                if (file.Source.Contains (Path.AltDirectorySeparatorChar) || file.Destination.Contains (Path.AltDirectorySeparatorChar))
-                    throw new InvalidOperationException ("DERP");
+            if (Path.AltDirectorySeparatorChar != Path.DirectorySeparatorChar) {
+                foreach (var file in source)
+                    if (file.Source.Contains (Path.AltDirectorySeparatorChar) || file.Destination.Contains (Path.AltDirectorySeparatorChar))
+                        throw new InvalidOperationException ("DERP");
```
alanmcgovern, line 255: I think the 'DERP' is aplaceholder which i meant to review/remove before merging the change. Who knows :p 

+1 on your change, but also... i  probably should fix/remove this exception so it's not just a DERP

alanmcgovern: I should probably just normalise values instead of throwing

IsaMorphic: > I should probably just normalise values instead of throwing

Absolutely! Will do.

alanmcgovern (approved):

src/MonoTorrent.Client/MonoTorrent/TorrentCreator.cs
alanmcgovern: https://github.com/alanmcgovern/monotorrent/commit/da09dd34b55120ddcb793258274651aa2462141b

That should be all that's needed to fix this properly! I'll see if i can configure CI to run on linux too.

Issue #698, opened 2024-09-13
Download Speed Gradually Drops to 0


kaedei:
In the latest version, if multiple download tasks (e.g., >5) are added simultaneously, the download speed for all tasks gradually drops to 0 after some time. This often happens after the first task is completed or when the remaining tasks are about 50% done.

I reviewed the recently updated code but haven't pinpointed the exact location or possible cause of the issue.

I tried rolling back the repository to versions from September 11, 2024, September 7, 2024, August 30, 2024, August 19, 2024, and August 10, 2024. I found that the batch download function works normally without the speed dropping to 0 only when I rolled back to the August 10, 2024 version (f9b120fe929353f5a27868dc4c38aac02c4b1434).

I looked through the nearby commits and suspect that https://github.com/alanmcgovern/monotorrent/commit/9a6db98878b74f26092f6ebd8940d914e3549a95 might be causing the issue, but I couldn't identify any problems in the code and am not sure if this judgment is correct.

alanmcgovern:
Interesting!

Out of curiosity, do you have logs from one of these sessions which indicates which messages were being sent when the connection was (eventually) dropped? If not, can you share one or two of the torrents where this is happening? Perhaps there's an edge case which I haven't hit yet which is causing one piece to be undownloadable, or repeatedly fail hash checks.

alanmcgovern:
@kaedei Also, could you try with the pre-release [3.0.3-alpha.unstable.rev0049](https://www.nuget.org/packages/MonoTorrent/3.0.3-alpha.unstable.rev0049) version in case that resolves the issue? I think i made a few more changes to upgradeable connections since that suspect commit landed after I identified some issues, so perhaps you're experiencing one of those?

kaedei:
> @kaedei Also, could you try with the pre-release [3.0.3-alpha.unstable.rev0049](https://www.nuget.org/packages/MonoTorrent/3.0.3-alpha.unstable.rev0049) version in case that resolves the issue? I think i made a few more changes to upgradeable connections since that suspect commit landed after I identified some issues, so perhaps you're experiencing one of those?

Thank you for the reminder. I created a demo project and tried using the latest version of the nuget package `3.0.3-alpha.unstable.rev0049` to download multiple torrent tasks. 
The code is available at: https://github.com/kaedei/monotorrent-303-speed-test

[...]

PR #723, opened 2025-11-02
fix connections leak

proff:
Probably fix this issue #698. At least in m case exactly same symptoms but on .net core


proff:
as you can see, the leak occurs under rare conditions for an ideal test environment. But quite real for the real environment.:
"Received handshake but protocol was unsupported"
"Could not receive a handshake from the peer"
ConnectionFailureReason.ConnectedToSelf
ConnectionFailureReason.Banned
ConnectionFailureReason.HandshakeFailed

alanmcgovern (approved):
nice catch!

alanmcgovern (approved):
SGTM! I'll add a test in a followup commit

PR #725, opened 2025-11-28
memory allocations optimization in MainLoop

proff:
using ConcurrentQueue instead of Queue recreate.

before
image
after
image
test scenario: same with #726

### DIFF
--- a/src/MainLoop.cs
+++ b/src/MainLoop.cs
@@ -29,6 +29,7 @@
 #nullable enable
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


src/MainLoop.cs
alanmcgovern, line 89: I'm a little confused. There are no allocations in the current codepath so i'm not sure what your profiling is showing. The current implementation allocates two queues and simply swaps them while the lock is held. Nothing is reallocated.

Changing to concurrent queue shouldn't affect total allocations.

alanmcgovern:
I'll close this for now, but if you remember what you were optimising and why this helps please do reopen it.

PR #726, opened 2025-12-08
remove redundant context switches

proff:
it is redundant to switch synchronization context to just switch again
image
performance test scenario: call GetLength with different thread and call count
image


alanmcgovern:
It is redundant, but it makes reasoning about IPieceWriter implementations much simpler. The invariant is "all `IPieceWriter` methods are invoked from the same thread, so implementations don't have to worry about multiple methods, or the same method, being invoked concurrently".

This invariant is important for Read/Write/Move as those APIs have caching.

PR #727, opened 2025-12-12
memory utilization optimization in BitFieldData

proff:
1.
image
2.
image
Most common BitField states are AllFalse or AllTrue. It's redundant to store per bit data in these states


alanmcgovern:
That's a neat idea! I just merged in the latest master. Everything still looks good so i'll merge this!

Thanks!

src/MonoTorrent/BitFieldData.cs
```diff
@@ -300,6 +341,13 @@ public int CountTrue (ReadOnlyBitField selector)
             if (selector.Length != Length)
                 throw new ArgumentException ("The selector should be the same length as this bitfield", nameof (selector));
 
+            if (selector.AllFalse || AllFalse)
+                return 0;
+            if (selector.AllTrue)
+                return TrueCount;
+            if (AllTrue)
+                return selector.TrueCount;
```
alanmcgovern, line 349: +1

alanmcgovern:
piece_picking

One downside of this is that piecepicking now re-allocates lots of uint64 arrays as the purpose of 'rarest first' picking is typically to initialise to an 'all true' bitfield (for the case where you're downloading from a seeder) and then keeping AND'ing in leecher bitfields until you have no bits set to true. This is how I generate the rarest set anyway :p 

The larger the torrent the more allocations there'll be. The cost of this could be amortised if BitField's had a cache pool for the underlying `ulong[]`. Not convinced that's worth it though :p

PR #728, opened 2025-12-13
redundant check of files exist when root directory is not exists

proff:
redundant check of files exist when root directory is not exists


alanmcgovern:
I think the majority of this form of issue *may* have been resolved already. Either way, this PR has a few dozen unrelated commits in it and also doesn't apply, so if there's still a remaining issue do let me know! Thanks, and apologies for not being able to do something with this PR specifically.

PR #729, opened 2025-12-14
memory allocations optimization in ListenManager

proff:
use ImmutableList instead of copy-on-change array (legacy ImmutableList implementation)
image
ListenManager.Add is inlined into ClientEngine.Register, so in profiler allocations in <Register>d_121.MoveNext


src/MonoTorrent.Client/MonoTorrent.Connections.Peer.Encryption/EncryptorFactory.cs
```diff
@@ -55,7 +55,7 @@ public EncryptorResult (IEncryption decryptor, IEncryption encryptor, HandshakeM
             }
         }
 
-        internal static async ReusableTask<EncryptorResult> CheckIncomingConnectionAsync (IPeerConnection connection, IList<EncryptionType> allowedEncryption, InfoHash[] sKeys, Factories factories, TimeSpan timeout)
+        internal static async ReusableTask<EncryptorResult> CheckIncomingConnectionAsync (IPeerConnection connection, IList<EncryptionType> allowedEncryption, IEnumerable<InfoHash> sKeys, Factories factories, TimeSpan timeout)
```
alanmcgovern:
```suggestion
        internal static async ReusableTask<EncryptorResult> CheckIncomingConnectionAsync (IPeerConnection connection, IList<EncryptionType> allowedEncryption, ImmutableList<InfoHash> sKeys, Factories factories, TimeSpan timeout)
```

src/MonoTorrent.Client/MonoTorrent.Connections.Peer.Encryption/EncryptorFactory.cs
```diff
@@ -65,7 +65,7 @@ internal static async ReusableTask<EncryptorResult> CheckIncomingConnectionAsync
             return await DoCheckIncomingConnectionAsync (connection, allowedEncryption, sKeys, factories).ConfigureAwait (false);
         }
 
-        static async ReusableTask<EncryptorResult> DoCheckIncomingConnectionAsync (IPeerConnection connection, IList<EncryptionType> preferredEncryption, InfoHash[] sKeys, Factories factories)
+        static async ReusableTask<EncryptorResult> DoCheckIncomingConnectionAsync (IPeerConnection connection, IList<EncryptionType> preferredEncryption, IEnumerable<InfoHash> sKeys, Factories factories)
```
alanmcgovern:
```suggestion
        static async ReusableTask<EncryptorResult> DoCheckIncomingConnectionAsync (IPeerConnection connection, IList<EncryptionType> preferredEncryption, ImmutableList<InfoHash> sKeys, Factories factories)
```

src/MonoTorrent.Client/MonoTorrent.Connections.Peer.Encryption/PeerBEncryption.cs
```diff
@@ -45,9 +45,9 @@ sealed class PeerBEncryption : EncryptedSocket
     {
         public byte[] InitialData { get; private set; }
 
-        InfoHash[] PossibleSKEYs { get; }
+        IEnumerable<InfoHash> PossibleSKEYs { get; }
```
alanmcgovern:
```suggestion
        ImmutableList<InfoHash> PossibleSKEYs { get; }
```

src/MonoTorrent.Client/MonoTorrent.Connections.Peer.Encryption/PeerBEncryption.cs
```diff
@@ -45,9 +45,9 @@ sealed class PeerBEncryption : EncryptedSocket
...
 
-        public PeerBEncryption (Factories factories, InfoHash[] possibleSKEYs, IList<EncryptionType> allowedEncryption)
+        public PeerBEncryption (Factories factories, IEnumerable<InfoHash> possibleSKEYs, IList<EncryptionType> allowedEncryption)
```
alanmcgovern:
```suggestion
        public PeerBEncryption (Factories factories, ImmutableList<InfoHash> possibleSKEYs, IList<EncryptionType> allowedEncryption)
```

alanmcgovern (commented):
Pass the ImmutableList by it's actual type

    These APIs are internal, so maintain the strong typing for this
    parameter. This means when the type is put in a 'foreach' loop it can
    use the non-boxing 'GetEnumerator' method and have zero/low allocations.

alanmcgovern:
Good call! It's one of the reasons i wanted to just move to a newer .NET version.

alanmcgovern:
My only question is - what are you doing to generate gigs of data via this API? How many concurrent torrents are you loading?

proff:
in all my MRs test scenario is: load and add about 15000 torrents.

alanmcgovern:
Ah interesting. Out of curiosity - how good/bad is the experience downloading/seeding using 15k torrents? I've only ever tried with hundreds

proff:
@alanmcgovern with real seeding and downloading i tried maximum about 3k seeding and maximum tens downloading. They are mostly inactive (waiting for requests), so it works normally. But in real production needs a lazy loading and unloading system (an api that allows you to check if an unknown torrent is available after receiving a request, and if it is, to download it).

DHT is completely unsuitable for such a large number of torrents and completely stops working, consuming one logical CPU at 100%. I fixed the perpetual loop that leads to this, but I forgot to create a PR (I'll create it later). However, it's better to turn it on anyway, because there's a lot of extra traffic)

alanmcgovern:
I know what the dht loop probably is. The spec suggests using two byte tokens for messages. That's 65k unique in-flight messages. With 5000 active torrents that'd allow for at most 13 in flight messages per torrent before there are collisions. 

It's quite likely that limit is being hit as I don't think Dht announces/searches are throttled in the same way as HTTP. Also DHT announces are 1 -> many in terms of requests. In order to complete a single bit torrent announce dozens of dht messages are sent to identify the closest peers to the infohash, and then those closest peers each get an announce message. 

I suspect in addition to throttling I should just move to 3 or 4 byte identifiers

alanmcgovern:
https://github.com/alanmcgovern/monotorrent/pull/738 for the DHT 100% issue. That should cover at least one cause.

alanmcgovern:
@proff out of curiosity could you run your benchmark with the latest code in master. If you see any API breakages let me know what they are. 

If you see a performance or memory improvement can you also give me a before/after comparison? I'm expecting you'll see substantial improvements in performance. 

Also, just in case you don't know, if you are loading 1000s of torrents it is safe to call ClientEngine.AddAsync multiple times concurrently. I'll document that as such. It greatly helped me when I used Paralell.ForEach to load 15k torrents with concurrency of 50 when benchmarking and developing these changes

PR #733, opened 2026-01-11
Fix AllFilesCorrectLength bug - torrents not marked complete after 100% downloaded

uponatime2019:
When a torrent completes downloading (Bitfield.AllTrue), RefreshAllFilesCorrectLengthAsync was not being called, so AllFilesCorrectLength remained false from startup/hash check. This caused torrents to reach 100% but never be marked as Complete.

This fix adds the RefreshAllFilesCorrectLengthAsync call in DownloadMode.UpdateSeedingDownloadingState when Bitfield.AllTrue, ensuring file lengths are verified before checking Complete status.

### DIFF
--- a/src/MonoTorrent.Client/MonoTorrent.Client.Modes/DownloadMode.cs
+++ b/src/MonoTorrent.Client/MonoTorrent.Client.Modes/DownloadMode.cs
@@ -90,6 +90,10 @@ namespace MonoTorrent.Client.Modes
         internal async Task UpdateSeedingDownloadingState ()
         {
             //If download is fully complete, set state to 'Seeding' and send an announce to the tracker.
+            if (Manager.Bitfield.AllTrue) {
+                await Manager.RefreshAllFilesCorrectLengthAsync ();
+            }
+
             if (Manager.Complete && state == TorrentState.Downloading) {
                 state = TorrentState.Seeding;
                 await Task.WhenAll (


src/MonoTorrent.Client/MonoTorrent.Client.Modes/DownloadMode.cs
```diff
@@ -90,6 +90,10 @@ public override void Tick (int counter)
         internal async Task UpdateSeedingDownloadingState ()
         {
             //If download is fully complete, set state to 'Seeding' and send an announce to the tracker.
+            if (Manager.Bitfield.AllTrue) {
+                await Manager.RefreshAllFilesCorrectLengthAsync ();
```
alanmcgovern, line 94: SGTM - makes sense.

I just want to add a test for that

alanmcgovern: I wrote a longer message but lost it when i was switching between tabs. Thanks github!

tl;dr I'm a little confused as to why this is necessary. `StartingMode` ensures all zero length files are created and that files which are 'too long' are truncated. This happens for all files which are not marked 'DoNotDownload'.

Once that logic runs, then `RefreshAllFilesCorrectLengthAsync` should always return true as all zero length files will exist *and* none will be too long.

The only edge case i can think of offhand is that `RefreshAllFilesCorrectLengthAsync` operates on all files, but CreateOrTruncate` operates on files which are not marked `DoNotDownload`, so files which are marked `DoNotDownload` won't exist and/or won't be the right length.

iow - i think this stuff needs to go away and be replaced with a `CreateOrTruncate` invocation whenever a torrent file changes priority. Or something like that?

alanmcgovern: Also something does need to ensure that zero length files have their bitfield set to 'true' when they exist on disk.

alanmcgovern:
@uponatime2019 Does this fix the issue you experienced?

https://github.com/alanmcgovern/monotorrent/pull/736

alanmcgovern:
I believe the underlying issue has been resolved in master now!  LMK if that isn't the case! Thanks for reporting this!

PR #738, opened 2026-05-21
Use both destination endpoint and DHT message ID to identify responses

alanmcgovern:
This should be better than simply increasing the transaction id size from 2 bytes to 4 bytes.

In the prior setup there is a chance that if there are hundreds of active torrents, each of which are announcing to the DHT engine, then all 65k unique transaction IDs could be consumed. Even if that didn't occur, the CPU cost of looking for an unused ID would be high. By further filtering by destination IP there's a stronger upper bound on the total number of scans needed to find an unused transaction ID.


proff:
@alanmcgovern 
that's how I fixed it, so I won't do a PR. However, I also took a precautionary measure: if a miracle happens and we're still in the loop after 65536 iterations, it's clear that it's pointless to continue, as we'll never break out of the loop. This is because we've completely blocked the DHT, and the identifier will never be freed.

PR #741, opened 2026-05-29
Allocation free parser

alanmcgovern:


src/MonoTorrent.Dht/MonoTorrent.Dht.Messages/Efficient.cs
```diff
@@ -0,0 +1,788 @@
...
+            ReadOnlyMemory<byte> values,
+            ReadOnlyMemory<byte> token)
+        {
+            this.nodes = nodes;
+            this.values = values;
+            this.token = token;
+        }
+
+        public bool HasNodes => !nodes.IsEmpty;
+        public bool HasValues => !values.IsEmpty;
```
alanmcgovern, line 148: maybe also assert that the lists/bencodedstrings aren't empty?

A list should have a span whose length is > 2. A string can only be empty if it's length is zero, which means `0:`, so the span must also be greater than 2.

alanmcgovern:
Way too many inter-related changes ended up in here. Let's see if i can split some out.

PR #747, opened 2026-06-08
Convert messages to `readonly ref struct` and discard the custom pooling

alanmcgovern:
BitTorrent messages are ephemeral. There can be between 10k and 100k generated per second depending on the datarate of them per second so in the past it was kinda useful from a memory perspective to pool them rather than allocating and throwing away.

With modern Span primitives there's no need to do that. 

If a message needs to be sent, a thin method can simply write the bytes directly to a Memory<T>, which in turn can be passed straight to the NetworkIO methods which, in turn, can encrypt in place before sending.

Similarly, when we use a buffer to the receive a message from a remote peer, we can wrap that Memory<T> in a simple struct which reads the underlying data using BinaryPrimitives.ReadUInt32BigEndian (etc). This gives a very simple model to operate against, removes complicated easy-to-screw-up lifespan management where buffers and message objects had differing lifespans, and gives one very simple model - There's a `Memory<T>` for each message and a releaser, that's it.


src/Samples/SampleClient/main.cs
```diff
@@ -15,6 +15,7 @@ class MainClass
     {
         static void Main (string[] args)
         {
+            new CompactEndPoint ();
```
alanmcgovern:
```suggestion
```

PR #749, opened 2026-06-13
Improved DHT bootstrap

alanmcgovern:
New bootstrapping approach:
1) Query the bootstrap nodes/routers for nodes close to the local ID.
2) Ping each of these to see if they're alive. If they respond they'll be inserted into the routing table. Up to 32 nodes will be stored in the first bucket. If the bucket splits then each will be allowed store 16. Further splits will use the standard 8 node limit.
3) Pre-split the routing table so it has at least 20 buckets to ensure there is reasonable diversity across the id space. The up-to-32 initial nodes will distribute appropriately.
4) Querying each bucket for a random node in each buckets range.

After the initial pings complete, allow the DHT table to be marked as 'healthy' or 'initialised' as soon as we have sufficient nodes in the table. Further background/non-blocking initialisation may continue until the full initialisation has completed, however the table is ready for querying once we have a bunch of healthy nodes.

This gets the table populated in far less time, the boosting of capacity during the initial bucket splits allows better diversity during bootstrap.

### DIFF
--- a/src/MonoTorrent.Dht/MonoTorrent.Dht/MessageLoop.cs
+++ b/src/MonoTorrent.Dht/MonoTorrent.Dht/MessageLoop.cs
@@ -40,7 +40,6 @@ using MonoTorrent.BEncoding;
 using MonoTorrent.Connections;
 using MonoTorrent.Connections.Dht;
 using MonoTorrent.Dht.Messages;
-using MonoTorrent.Dht.Messages.Efficient;
 using MonoTorrent.Logging;
 
 using ReusableTasks;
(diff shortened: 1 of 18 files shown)


src/MonoTorrent.Dht/MonoTorrent.Dht/MessageLoop.cs
```diff
@@ -126,7 +125,7 @@ public MessageLoop (DhtEngine engine, TransferMonitor monitor)
             Listener = new NullDhtListener ();
             ReceiveQueue = new Queue<KeyValuePair<CompactEndPoint, ReadOnlyMemory<byte>>> ();
             SendQueue = new Queue<SendDetails> ();
-            Timeout = TimeSpan.FromSeconds (10);
+            Timeout = TimeSpan.FromSeconds (2);
```
alanmcgovern, line 128: we do 3 retries of a message during init, so 2 seconds wait between retries for 6 seconds total should be fine. If a node isn't responding, then it never will.

src/MonoTorrent.Dht/MonoTorrent.Dht/MessageLoop.cs
```diff
@@ -198,7 +197,7 @@ void RaiseMessageSent (Node node, CompactEndPoint endpoint, ReadOnlyMemory<byte>
 
         async Task SendMessages ()
         {
-            for (int i = 0; i < 50 && SendQueue.Count > 0; i++) {
+            for (int i = 0; i < 150 && SendQueue.Count > 0; i++) {
```
alanmcgovern, line 200: This should be kilobytes per second at worst

PR #755, opened 2026-06-28
Implement UtpPacket as per library norms

alanmcgovern:
The underlying data will be a Memory<byte> so the timestamps can be written right before sending so having a 'ReadOnlyMemory' version for sending is not particularly useful.

By wrapping a Memory<byte>, as like the rest of the library, the UtpPacket itself can be a simple struct holding the memory. The underlying Memory<byte> will come from a pool which is sized at MTU bytes. Some care will be needed to slice the underlying Memory prior to sending it across the socket to ensure it is sliced to the actual size of the underlying data.

The intended flow will be:
- Engine calls 'SendAsync' with a buffer of data.
- The UTP connection will reformat this data into N packets and enqueue them for sending.
- The 'SendAsync' will be marked as 'done' so the engine knows more data than be sent.
- If the send queue is full, then the utp connection will not allow subsequent 'SendAsync' calls to return until queue space has freed up and the latest message has been reformatted and enqueued.


src/Samples/SampleClient/main.cs
```diff
@@ -51,15 +51,18 @@ static async Task MainAsync (string[] args, CancellationToken token)
                 // it's cache directory. Otherwise the metadata will be downloaded and stored in the cache directory
                 // so it can be reloaded later.
                 AutoSaveLoadMagnetLinkMetadata = true,
-
+                AllowedTransports =ImmutableArray.Create (PeerTransport.Utp),
```
alanmcgovern:
```suggestion
```

src/MonoTorrent/MonoTorrent/ByteBufferPool.cs
```diff
@@ -37,6 +37,7 @@ public partial class ByteBufferPool
     {
         const int AllocateDelta = 8;
 
+
```
alanmcgovern:
```suggestion
```

src/MonoTorrent/MonoTorrent/MemoryPool.cs
```diff
@@ -57,6 +57,7 @@ public ByteBufferPool.Releaser Rent (int capacity, out Memory<byte> memory)
                 : capacity <= LargeMessageBufferSize
                     ? LargeMessageBuffers
                     : MassiveBuffers;
+
```
alanmcgovern:
```suggestion
```

src/Samples/SampleClient/main.cs
```diff
@@ -51,15 +51,17 @@ static async Task MainAsync (string[] args, CancellationToken token)
...
                 ListenEndPoints = new Dictionary<string, IPEndPoint> {
                     { "ipv4", new IPEndPoint (IPAddress.Any, 55123) },
                     { "ipv6", new IPEndPoint (IPAddress.IPv6Any, 55123) }
                 }.ToImmutableDictionary (),
 
                 // Use a fixed port for DHT communications for testing purposes. Production usages should use a random port, 0, if possible.
-                DhtEndPoint = new IPEndPoint (IPAddress.Any, 55123),
+                UdpListenEndPoints = new Dictionary<string, IPEndPoint> {
+                    { "ipv4", new IPEndPoint (IPAddress.Any, 55124) },
+                    { "ipv6", new IPEndPoint (IPAddress.IPv6Any, 55124) }
```
alanmcgovern:
```suggestion
                    { "ipv4", new IPEndPoint (IPAddress.Any, 55123) },
                    { "ipv6", new IPEndPoint (IPAddress.IPv6Any, 55123) }
```

alanmcgovern:
ucat.exe was improperly added to the repo. Strip it out of history before merging

PR #756, opened 2026-07-12
Enable creating/buffering utp packets

alanmcgovern:
Refactor the buffer pool to support setting the buffer size. Use this to create a 1500 byte udp packet buffer.


alanmcgovern:
I handled it a different way

